PR #9, opened 2024-07-19
Update SearchController.php

vldmr-k:
Remove debug information


loevgaard:
Hi, @vldmr-k 

The plugin is still in development stage, so the code is not ready for production yet. Also, as you may have noticed the build isn't green, and I will never release something that isn't green :)

PR #11, opened 2024-08-20
MS-21 Prices indexing fix

Zales0123:
That's the first thing we need to do to make sorting work. Product prices were wrongly indexed taking into account how they're displayed in standard Sylius - which is not the lowest price but the price from the first variant. For now, products were indexed with the lowest price and it looked like things were not working :dancer:

The result of the fix: 

https://github.com/user-attachments/assets/b6205494-d0ac-4744-9f76-e774928d0382

I believe we should stick to the default Sylius approach, to make this plugin work on the standard Sylius UI. The service for price indexing is extracted, so it can be easily overridden.

@loevgaard if you think we should instead change the way how prices are displayed in the twig templates - let me know 🖖

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -24,6 +24,7 @@
     </plugins>
     <forbiddenFunctions>
         <function name="dd"/>
+        <function name="dump"/>
         <function name="exit"/>
         <function name="var_dump"/>
         <function name="print_r"/>
--- a/psalm.xml
+++ b/psalm.xml
@@ -24,7 +24,6 @@
     </plugins>
     <forbiddenFunctions>
         <function name="dd"/>
-        <function name="dump"/>
         <function name="exit"/>
         <function name="var_dump"/>
         <function name="print_r"/>
--- a/src/DataMapper/Product/PriceDataMapper.php
+++ b/src/DataMapper/Product/PriceDataMapper.php
@@ -5,10 +5,9 @@ declare(strict_types=1);
 namespace Setono\SyliusMeilisearchPlugin\DataMapper\Product;
 
 use Setono\SyliusMeilisearchPlugin\DataMapper\DataMapperInterface;
-use Setono\SyliusMeilisearchPlugin\DataMapper\Product\Provider\ProductIndexedPricesProviderInterface;
+use Setono\SyliusMeilisearchPlugin\DataMapper\Product\Provider\ProductPricesProviderInterface;
 use Setono\SyliusMeilisearchPlugin\Document\Document;
 use Setono\SyliusMeilisearchPlugin\Document\Product as ProductDocument;
-use function Setono\SyliusMeilisearchPlugin\formatAmount;
 use Setono\SyliusMeilisearchPlugin\Model\IndexableInterface;
 use Setono\SyliusMeilisearchPlugin\Provider\IndexScope\IndexScope;
 use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
@@ -16,13 +15,14 @@ use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Currency\Converter\CurrencyConverterInterface;
 use Webmozart\Assert\Assert;
+use function Setono\SyliusMeilisearchPlugin\formatAmount;
 
 final class PriceDataMapper implements DataMapperInterface
 {
     public function __construct(
         private readonly ChannelRepositoryInterface $channelRepository,
         private readonly CurrencyConverterInterface $currencyConverter,
-        private readonly ProductIndexedPricesProviderInterface $productIndexedPricesProvider,
+        private readonly ProductPricesProviderInterface $productIndexedPricesProvider,
     ) {
     }
 
--- a/src/DataMapper/Product/PriceDataMapper.php
+++ b/src/DataMapper/Product/PriceDataMapper.php
@@ -8,6 +8,7 @@ use Setono\SyliusMeilisearchPlugin\DataMapper\DataMapperInterface;
 use Setono\SyliusMeilisearchPlugin\DataMapper\Product\Provider\ProductIndexedPricesProviderInterface;
 use Setono\SyliusMeilisearchPlugin\Document\Document;
 use Setono\SyliusMeilisearchPlugin\Document\Product as ProductDocument;
+use function Setono\SyliusMeilisearchPlugin\formatAmount;
 use Setono\SyliusMeilisearchPlugin\Model\IndexableInterface;
 use Setono\SyliusMeilisearchPlugin\Provider\IndexScope\IndexScope;
 use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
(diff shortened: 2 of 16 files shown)


psalm.xml
```diff
@@ -24,7 +24,6 @@
     </plugins>
     <forbiddenFunctions>
         <function name="dd"/>
-        <function name="dump"/>
```
Zales0123, line 27: I remove it for now as we actually have some dumps in the code now 🚀 Should be added later.

tests/Application/config/packages/setono_sylius_meilisearch.yaml
```diff
@@ -8,4 +8,4 @@ setono_sylius_meilisearch:
             entities: [ 'Setono\SyliusMeilisearchPlugin\Tests\Application\Entity\Taxon' ]
     search:
         index: products
-        hits_per_page: 2
+        hits_per_page: 30
```
Zales0123, line 11: I've changed it to have more products to be sorted. It will probably be configurable from the UI in the future, I suppose 🖖

psalm.xml
Zales0123: Or, I will just remove those dumps 🔥

loevgaard:
Oh, yeah, good point!

src/DataMapper/Product/PriceDataMapper.php
```diff
@@ -13,18 +14,15 @@
 use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
 use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\ProductInterface;
-use Sylius\Component\Core\Model\ProductVariantInterface;
 use Sylius\Component\Currency\Converter\CurrencyConverterInterface;
 use Webmozart\Assert\Assert;
 
-/**
- * This data mapper maps prices on product documents
- */
```
loevgaard, line 22: Is it so obvious? 😂

src/DataMapper/Product/Provider/ProductIndexedPricesProvider.php
```diff
@@ -0,0 +1,33 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\DataMapper\Product\Provider;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\ProductInterface;
+use Sylius\Component\Core\Model\ProductVariantInterface;
+
+final class ProductIndexedPricesProvider implements ProductIndexedPricesProviderInterface
```
loevgaard, line 11: I don't like this name, what's the reasoning? Could it just be `ProductVariantPricesProvider` or `ProductPricesProvider`?

src/DataMapper/Product/Provider/ProductIndexedPricesProvider.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /** We need to provide the price from the first valid variant, to reflect the default Sylius UI behavior */
+    public function getPricesForChannel(ProductInterface $product, ChannelInterface $channel): ProductPrices
```
loevgaard, line 14: Shouldn't we use the default Sylius service to get the variant also? I.e. the `ProductVariantResolverInterface`

src/DataMapper/Product/Provider/ProductIndexedPricesProvider.php
```diff
@@ -0,0 +1,33 @@
...
+        $enabledVariants = $product->getEnabledVariants();
+        if ($enabledVariants->isEmpty()) {
+            return new ProductPrices();
+        }
+
+        /** @var ProductVariantInterface $firstEnabledVariant */
+        $firstEnabledVariant = $enabledVariants->first();
+        $firstChannelPricing = $firstEnabledVariant->getChannelPricingForChannel($channel);
+        if (null === $firstChannelPricing) {
+            return new ProductPrices();
```
loevgaard, line 25: I believe this is wrong. Shouldn't it be an exception or return a special 'null object' of the `ProductPrices`?

src/DataMapper/Product/Provider/ProductPrices.php
```diff
@@ -0,0 +1,14 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\DataMapper\Product\Provider;
+
+final class ProductPrices
+{
+    public function __construct(
+        public readonly ?int $price = null,
+        public readonly ?int $originalPrice = null,
```
loevgaard, line 11: It really doesn't matter, but a value object like this shouldn't have both properties nullable imo. But again, it doesn't matter in our use case

src/DataMapper/Product/PriceDataMapper.php
```diff
@@ -5,26 +5,24 @@
...
 final class PriceDataMapper implements DataMapperInterface
 {
     public function __construct(
         private readonly ChannelRepositoryInterface $channelRepository,
         private readonly CurrencyConverterInterface $currencyConverter,
+        private readonly ProductPricesProviderInterface $productIndexedPricesProvider,
```
loevgaard, line 25:
```suggestion
        private readonly ProductPricesProviderInterface $productPricesProvider,
```

src/DataMapper/Product/Provider/ProductPrices.php
```diff
@@ -0,0 +1,14 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\DataMapper\Product\Provider;
+
+final class ProductPrices
+{
+    public function __construct(
+        public readonly int $price,
+        public readonly int $originalPrice,
```
loevgaard, line 11: I guess the original price should be nullable. Maybe I was a bit confusing on my first comment :)

tests/DataMapper/Product/Provider/ProductPricesProviderTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $channelPricing = new ChannelPricing();
+        $channelPricing->setChannelCode('MY_CHANNEL');
+        $channelPricing->setPrice(1000);
+        $channelPricing->setOriginalPrice(2000);
+
+        $productVariant = new ProductVariant();
+        $productVariant->addChannelPricing($channelPricing);
+
+        $this->productVariantResolver->method('getVariant')->willReturn($productVariant);
```
loevgaard, line 41: If you want phpspec like prophecies, you can use this package: https://packagist.org/packages/phpspec/prophecy-phpunit

src/DataMapper/Product/Provider/ProductPrices.php
jakubtobiasz: Yup, I know, I started a meeting before I had fixed that :D

tests/DataMapper/Product/Provider/ProductPricesProviderTest.php
jakubtobiasz: I'll stick to PHPUnit Mock unless you decide we should standardize it and use prophecy :D

jakubtobiasz:
@loevgaard everything fixed, BCC and Mutation tests are failing, but I believe it's not related to this PR :).

src/DataMapper/Product/PriceDataMapper.php
Zales0123: I seems so, it's a **PriceDataMapper**, what else can it do? 😄

loevgaard:
Great! Thank you, @jakubtobiasz and @Zales0123 🍾

PR #12, opened 2024-08-21
Add sorting of results

Zales0123:
Based on https://github.com/Setono/sylius-meilisearch-plugin/pull/11

Added ability to sort the search results by 4 criteria:
- Cheapest first (lowest price)
- Biggest discount (biggest difference between price and originalPrice)
- Newest first (by creation date)
- Relevance (set by Meilisearch)

Here is a video showing the result:

https://github.com/user-attachments/assets/399078b6-a9db-4992-bd5b-33408f8d794c

In this PR I've also added some functional tests, that require running Meilisearch locally - so they're useful for the development (we don't have to click on the website :dancer:) but are not yet runnable on CI (we could probably make it happen). A bunch of refactoring is also there 🚀 

To-do:
- [x] do not require to click "Update filter" after changing the sorting (do it automatically)


src/Engine/SearchEngine.php
```diff
@@ -0,0 +1,46 @@
...
+
+use Meilisearch\Client;
+use Meilisearch\Search\SearchResult;
+use Setono\SyliusMeilisearchPlugin\Config\Index;
+use Setono\SyliusMeilisearchPlugin\Document\Metadata\Facet;
+use Setono\SyliusMeilisearchPlugin\Document\Metadata\MetadataFactoryInterface;
+use Setono\SyliusMeilisearchPlugin\Meilisearch\Builder\FilterBuilderInterface;
+use Setono\SyliusMeilisearchPlugin\Resolver\IndexName\IndexNameResolverInterface;
+
+final class SearchEngine
```
loevgaard, line 15: We need an interface.

I absolutely love the refactoring!

loevgaard:
Can you merge #11 and rebase? It's a bit easier to review then :D

composer.json
```diff
@@ -108,7 +108,7 @@
         "analyse": "psalm",
         "check-style": "ecs check",
         "fix-style": "ecs check --fix",
-        "phpunit": "phpunit",
+        "phpunit": "phpunit --exclude-group=functional",
```
loevgaard, line 111: I think it's easier to use suites, don't you? I think we did the same in the order edit plugin

src/Resources/config/services/conditional/search.xml
```diff
@@ -10,12 +10,16 @@
...
             <argument type="service" id="doctrine"/>
             <argument type="service" id="twig"/>
-            <argument type="service" id="setono_sylius_meilisearch.resolver.index_name"/>
-            <argument type="service" id="Meilisearch\Client"/>
-            <argument type="service" id="Setono\SyliusMeilisearchPlugin\Document\Metadata\MetadataFactoryInterface"/>
             <argument type="service" id="Setono\SyliusMeilisearchPlugin\Form\Builder\SearchFormBuilderInterface"/>
+            <argument type="service" id="Setono\SyliusMeilisearchPlugin\Engine\SearchEngine" />
+        </service>
+
+        <service id="Setono\SyliusMeilisearchPlugin\Engine\SearchEngine" public="true">
```
loevgaard, line 17:
```suggestion
        <service id="Setono\SyliusMeilisearchPlugin\Engine\SearchEngine">
```

loevgaard:
> do not require to click "Update filter" after changing the sorting (do it automatically)

Is that because you haven't run `php bin/console assets:install` maybe? Because it should submit automatically. See https://github.com/Setono/sylius-meilisearch-plugin/blob/master/src/Resources/public/js/search-manager.js#L58

jakubtobiasz:
> do not require to click "Update filter" after changing the sorting (do it automatically)

It works for me too

tests/Functional/SearchTest.php
```diff
@@ -0,0 +1,76 @@
...
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Tests\Functional;
+
+use Setono\SyliusMeilisearchPlugin\Engine\SearchEngine;
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+
+/** @group functional */
+final class SearchTest extends WebTestCase
```
loevgaard, line 12: Very nice to have tests! :D

tests/Functional/SearchTest.php
```diff
@@ -0,0 +1,76 @@
...
+            self::assertGreaterThanOrEqual($previousHit['price'], $hit['price']);
+            $previousKey = $key;
+        }
+    }
+
+    public function testItSortsSearchResultsByNewestDate(): void
+    {
+        /** @var SearchEngine $searchEngine */
+        $searchEngine = self::getContainer()->get(SearchEngine::class);
+        $result = $searchEngine->execute('jeans', ['sort' => 'createdAt:desc']);
```
loevgaard, line 59: This small thing `createdAt:desc` made me think about whether we should snake case properties inside Meilisearch? If there would ever be a problem with camel case? I am thinking about web servers ignoring case, browsers ignoring case or whatever. I don't know. Do any of you know? It will give a lot of other problems though :)

loevgaard (approved):

tests/Functional/SearchTest.php
jakubtobiasz: IMO it shouldn't make any troubles.

jakubtobiasz (approved):

PR #14, opened 2024-08-21
Optimize indexing

loevgaard:
This PR will split the indexing into batches dispatched on the command bus. This will speed up the indexing process proportionally with the number of workers.


src/DataProvider/DefaultDataProvider.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function getObjects(string $entity, Index $index): \Generator
+    {
+        $qb = $this
+            ->getManager($entity)
+            ->createQueryBuilder()
+            ->select('o')
+            ->from($entity, 'o');
+
+        $this->eventDispatcher->dispatch(new DataProviderEvent($entity, $index, $qb));
```
jakubtobiasz, line 34: I have mixed feelings about the event's name. The event name should reflect a something that happened in the past. So maybe something like `DataObjectProvided` or `PreDataObjectProvided` if you want allow to e.g. modify a $qb?

src/Indexer/IndexBuffer.php
jakubtobiasz, line 1: Where is it used?

src/DataProvider/DefaultDataProvider.php
loevgaard: Yeah, agreed :)

src/Indexer/IndexBuffer.php
loevgaard: In the `DefaultIndexer`

PR #16, opened 2024-08-22
Add autocomplete functionality for the search input

loevgaard:


src/Resources/views/autocomplete/javascript.html.twig
```diff
@@ -0,0 +1,125 @@
...
+                                    {
+                                        indexName: '{{ source.index }}',
+                                        query,
+                                    },
+                                ],
+                            })
+                        },
+                        templates: {
+                            item({ item, components, html }) {
+                                return html`<div class="aa-ItemWrapper">
```
loevgaard, line 54: This template should be moved to a twig template and rendered

src/Resources/views/autocomplete/javascript.html.twig
```diff
@@ -0,0 +1,125 @@
...
+                                                    />
+                                                </svg>
+                                            </button>
+                                        </div>
+                                    </div>
+                                </div>`;
+                            },
+                        },
+                        getItemUrl({ item }) {
+                            return item.url;
```
loevgaard, line 117: The property where the url is defined, should also be set on the `Source`. It should probably be nullable, and then only include the `getItemUrl` method is the property is not null

src/Resources/views/autocomplete/javascript.html.twig
```diff
@@ -0,0 +1,125 @@
+{% if ssm_autocomplete_enabled() %}
```
loevgaard, line 1: My idea for refactoring this file is to divide it into multiple files, so that it's easier for the user to pick and choose and override smaller parts.

Wdyt?

src/Resources/views/autocomplete/javascript.html.twig
```diff
@@ -0,0 +1,62 @@
...
+    {# @var configuration \Setono\SyliusMeilisearchPlugin\Meilisearch\Autocomplete\Configuration\Configuration #}
+    {% set configuration = ssm_autocomplete_configuration() %}
+    {% set templates = {
+        item: include('@SetonoSyliusMeilisearchPlugin/autocomplete/templates/item.html.twig'),
+    } %}
```
loevgaard, line 6: Should be saved on the `Source` object instead

src/Resources/public/css/algolia.autocomplete.css
Zales0123, line 1: algolia 😅

src/Twig/AutocompleteExtension.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Setono\SyliusMeilisearchPlugin\Twig;
+
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+final class AutocompleteExtension extends AbstractExtension
+{
+    /**
+     * @param bool $enabled Whether the autocomplete functionality is enabled or not
```
Zales0123, line 13: Pretty obvious, isn't it? 😄

src/Resources/views/autocomplete/templates/item.html.twig
```diff
@@ -0,0 +1,15 @@
+<div class="aa-ItemWrapper">
```
Zales0123, line 1: Shouldn't the item be clickable to be able to follow to the result? If it's not, the whole feature is quite useless 😅 

https://github.com/user-attachments/assets/61383fd2-d6ab-4d40-8671-8169851e6e7c

src/Resources/public/js/algolia.autocomplete.js
Zales0123, line 1: why to we need both, `algolia.autocomplete.js` and `meilisearch.autocomplete.js`?

src/Meilisearch/Autocomplete/Configuration/Source.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Setono\SyliusMeilisearchPlugin\Meilisearch\Autocomplete\Configuration;
+
+final class Source
+{
+    public function __construct(
+        public readonly string $id,
+        public readonly string $index,
+        /**
+         * The attribute that holds the URL on any given item
+         */
```
Zales0123, line 14: Imo the comment is not needed, but if you insist it to be here, I would at least make it one-liner 🖖

src/Meilisearch/Autocomplete/Configuration/Configuration.php
```diff
@@ -0,0 +1,27 @@
...
+        public readonly string $searchKey,
+
+        /**
+         * This is the javascript selector for the HTML element that will contain the autocomplete
+         */
+        public readonly string $container,
+
+        /**
+         * This is the placeholder text that will be displayed in the input field
+         */
```
Zales0123, line 23: Also here, both of the comments are redundant in my opinion :dancer:

src/Resources/public/css/algolia.autocomplete.css
loevgaard: Hehe, it is Algolias theme :D Meilisearch uses Algolias autocomplete :)

src/Resources/views/autocomplete/templates/item.html.twig
loevgaard: Yes, it should and it is on my computer 😂  I will figure it out!

src/Resources/public/js/algolia.autocomplete.js
loevgaard: Algolia provides the actual autocomplete library. Meilisearch then provides an adapter.

src/Meilisearch/Autocomplete/Configuration/Configuration.php
loevgaard: Aaaaaah, not quite. The name of the class is quite generic, so you wouldn't be completely sure unless you traversed some files. But I will make them one liners just for you <3

src/Meilisearch/Autocomplete/Configuration/Source.php
loevgaard: One-lined :D About comments, I believe more is more :D

src/Resources/views/autocomplete/templates/item.html.twig
loevgaard: Aaah, didn't notice you were clicking :D It works with the keyword. I will fix it! Thank you!

Zales0123 (approved):

PR #17, opened 2024-08-23
Refactor the Search Engine & Filter Builder

jakubtobiasz:
- Filter Builder has been split into smaller, specific filter focused services
- Search Engine has been refactored to use Multi Search to improve returned facet distribution


src/DataMapper/Product/Provider/ProductPricesProvider.php
```diff
@@ -8,20 +8,14 @@
 use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
-use Sylius\Component\Product\Resolver\ProductVariantResolverInterface;
 
 final class ProductPricesProvider implements ProductPricesProviderInterface
 {
-    public function __construct(
```
loevgaard, line 15: This shouldn't be removed because this is the default service for resolving a variant in Sylius

composer.json
```diff
@@ -88,6 +88,7 @@
     },
     "autoload-dev": {
         "psr-4": {
+            "TestApp\\": "tests/Application/src/",
```
loevgaard, line 91: What's the reasoning behind a new namespace?

src/DataMapper/Product/Provider/ProductPricesProvider.php
jakubtobiasz: But it causes issues on fixtures load. Additionally, the logic I used is almost the same as what this service does (the except is not using the repository).

composer.json
jakubtobiasz: I've created a `src` inside the test app, as I wanted to skip putting it directly under the `tests/Application`. The namespace will look `CurrentNamespace/src/FilterBuilder` so I decided to create a shorter version, as it doesn't impact the project itself.

src/DataMapper/Product/Provider/ProductPricesProvider.php
loevgaard: I created this PR: https://github.com/Setono/sylius-meilisearch-plugin/pull/18 specifically for this bug :)

composer.json
Zales0123: so, maybe, to have it consistent with other namespaces name it `Setono\\SyliusMeilisearchPlugin\\TestApp\\`? I know it's long :dancer: but who cares when we have autocomplete 🚀

src/Meilisearch/Builder/CompositeFilterBuilder.php
```diff
@@ -0,0 +1,35 @@
...
+
+final class CompositeFilterBuilder implements FilterBuilderInterface
+{
+    /** @var array<FilterBuilderInterface> */
+    private array $filterBuilders;
+
+    public function __construct(
+        iterable $filterBuilders,
+    ) {
+        $this->filterBuilders = $filterBuilders instanceof \Traversable ? iterator_to_array($filterBuilders) : $filterBuilders;
```
Zales0123, line 15: Is it even needed? \Traversable, iterable, array - all should be iterable in the same way 🖖

tests/Unit/Meilisearch/Builder/CompositeFilterBuilderTest.php
```diff
@@ -0,0 +1,114 @@
...
+        $brandFilterBuilder = new class() implements FilterBuilderInterface {
+            public function build(array $facets): string|array
+            {
+                $brandQuery = [];
+                /** @var string $size */
+                foreach ($facets['brand'] as $size) {
+                    $brandQuery[] = sprintf('brand = "%s"', $size);
+                }
+
+                return '(' . implode(' OR ', $brandQuery) . ')';
```
Zales0123, line 24: I suppose we can simplify the implementations of these anonymous classes, they're not relevant in the scope of this test

src/Engine/SearchEngine.php
Zales0123, line 1: I feel like this class is much less readable after the refactor than before, which should not be the case 😅 Let's maybe introduce another layer of abstraction, so this class would only gather the query and parameters, pass them to query builder (main query builder and subsearch query builders) and then execute the call on Meilisearch API? 

Let's start the discussion 🖖

src/Meilisearch/Builder/FilterBuilderInterface.php
```diff
@@ -6,5 +6,7 @@
 
 interface FilterBuilderInterface
 {
-    public function build(array $parameters): array;
+    public function build(array $facets): string|array;
```
Zales0123, line 9: We could potentially unify the return type, to always return array (no matter if with one or multiple elements)

src/DataMapper/Product/Provider/ProductPricesProvider.php
jakubtobiasz: Yup, reverted locally.

Zales0123:
Closing in favor of #22 - thank you @jakubtobiasz 🫡

PR #22, opened 2024-08-30
Multisearch + dynamic filters V2

Zales0123:
Replaces https://github.com/Setono/sylius-meilisearch-plugin/pull/17

In the beginning, I wanted to separate those concepts, as I already wrote to @loevgaard... but as I started the refactor and improvements on #17 it was eventually hard (and potentially time-consuming) to come back to the single search feature. Therefore, I decided to left them coupled to each other, I hope the PR is not too big (spoiler alert: it will be xD :dancer:)

The main things that were done in this PR:
- multi-search functionality (added already in https://github.com/Setono/sylius-meilisearch-plugin/pull/17)
- slight `SearchEngine` refactoring (see to-dos)
- dynamic filters - compared to implementation from the previous PR, filters are generated based on the `#[Facet]` attribute on the documented entity. The type of the filter is generic, based on the type of the property, which does not have to be defined in the attribute (it's taken from the property/method reflection data)
- dynamic filers form types (that also use the `#[Facet]` attribute)
- added a new filter for price + a corresponding filter form type (I don't know if it was required xD but I suppose it does not harm, as we approach facets in a generic way :rocket:)

The current functionality is shown in the video below:

https://github.com/user-attachments/assets/1b23bc1b-bc78-4df8-a742-369720e24063

What is left to be done:
- [ ] refactor `SearchEngine` fully (there are a couple of private methods there, each of them could and should be a separate service)
- [ ] simplify facets forms builders (I have a feeling they're a little bit overcomplicated(
- [ ] improve the price filter UI (it's hideous 🤮) 
- [ ] add some test cases to the `SearchTest`

🖖


src/Engine/SearchEngine.php
```diff
@@ -25,21 +27,105 @@ public function __construct(
 
     public function execute(?string $query, array $parameters = []): SearchResult
     {
-        $page = max(1, (int) ($parameters['p'] ?? 1));
-        $sort = (string) ($parameters['sort'] ?? '');
-
         $metadata = $this->metadataFactory->getMetadataFor($this->index->document);
+        $indexUid = $this->indexNameResolver->resolve($this->index);
+        /** @var array<string> $facetsNames */
```
loevgaard, line 32: This PHPDoc shouldn't be needed. It should be on the `getFacetableAttributeNames` instead

src/Engine/SearchEngine.php
```diff
@@ -25,21 +27,105 @@ public function __construct(
 
     public function execute(?string $query, array $parameters = []): SearchResult
     {
-        $page = max(1, (int) ($parameters['p'] ?? 1));
-        $sort = (string) ($parameters['sort'] ?? '');
-
         $metadata = $this->metadataFactory->getMetadataFor($this->index->document);
+        $indexUid = $this->indexNameResolver->resolve($this->index);
```
loevgaard, line 31:
```suggestion
        $indexName = $this->indexNameResolver->resolve($this->index);
```

I have used the term `indexName` in all other places.

src/Form/Builder/CheckboxFacetFormBuilder.php
```diff
@@ -4,13 +4,14 @@
...
 
+use Setono\SyliusMeilisearchPlugin\Document\Metadata\Facet;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\FormBuilderInterface;
 use function Symfony\Component\String\u;
 
 final class CheckboxFacetFormBuilder implements FacetFormBuilderInterface
 {
-    public function build(FormBuilderInterface $builder, string $name, array $values, array $stats = null): void
+    public function build(FormBuilderInterface $builder, string $name, array $values, Facet $facet, array $stats = null): void
```
loevgaard, line 14: The name could be fetched from the `Facet` object now, couldn't it?

src/Meilisearch/Builder/CompositeFilterBuilder.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Meilisearch\Builder;
+
+final class CompositeFilterBuilder implements FilterBuilderInterface
```
loevgaard, line 7: Rather use the `Setono\CompositeCompilerPass\CompositeService` that's used in several places in the plugin. Here are the docs: https://github.com/Setono/composite-compiler-pass

src/Form/Builder/CheckboxFacetFormBuilder.php
Zales0123: Indeed! 👌

src/Document/Metadata/Metadata.php
```diff
@@ -167,4 +167,13 @@ private static function resolveName(\ReflectionProperty|\ReflectionMethod $refle
 
         return null;
     }
+
+    private function getFacetType(\ReflectionProperty|\ReflectionMethod $attributesAware): string
```
loevgaard, line 171: The other methods are static for no particular reason. Just pointing it out :D I think I read sometime in some SF code that static methods was faster or had a smaller memory footprint. I have never googled it or fact checked it :D

src/Document/Metadata/MetadataInterface.php
```diff
@@ -23,7 +23,7 @@ public function getFilterableAttributes(): array;
     /**
      * Returns the names of the filterable attributes
      *
-     * @return list<string>
+     * @return array<string>
```
loevgaard, line 26: Why do you change it to something less specific?

src/Meilisearch/Filter/CompositeFilterBuilder.php
```diff
@@ -0,0 +1,22 @@
...
+
+/** @extends CompositeService<FilterBuilderInterface> */
+final class CompositeFilterBuilder extends CompositeService implements FilterBuilderInterface
+{
+    public function build(array $facets, array $facetsValues): array
+    {
+        $filters = [];
+
+        foreach ($this->services as $filterBuilder) {
+            $filters = array_merge($filters, $filterBuilder->build($facets, $facetsValues));
```
loevgaard, line 17: I think the PHP plugin in PHPStorm warns against this. You can read about it here: https://www.exakat.io/en/speeding-up-array_merge/

It should be `$filters[] = $filterBuilder->build($facets, $facetsValues)`

and then outside the loop:

`$filters = array_merge(...$filters);`

src/Meilisearch/Query/MainQueryBuilder.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Meilisearch\Query;
+
+use Meilisearch\Contracts\SearchQuery;
+
+final class MainQueryBuilder implements MainQueryBuilderInterface
```
loevgaard, line 9: I feel like having both a main query builder and a sub query builder is a bit weird, but I can't put my finger on it... What's your feeling about it?

src/Meilisearch/Query/SearchQueryBuilder.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Setono\SyliusMeilisearchPlugin\Meilisearch\Query;
+
+use Meilisearch\Contracts\SearchQuery;
+
+final class SearchQueryBuilder implements SearchQueryBuilderInterface
+{
+    public function build(string $indexName, string $query, array $facets, array $filter): SearchQuery
+    {
+        return (new SearchQuery())
+            ->setIndexUid($indexName)
```
loevgaard, line 14: Ah, Meilisearch actually calls it 'uid'... Hmm... I always felt that 'index name' was a bit ambiguous, but couldn't think of anything better. Maybe we should rename it in another PR

src/Meilisearch/Query/SearchQueryBuilderInterface.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Meilisearch\Contracts\SearchQuery;
+
+interface SearchQueryBuilderInterface
+{
+    /**
+     * @param array<string> $facets
+     * @param array<string, mixed> $filter
+     */
+    public function build(string $indexName, string $query, array $facets, array $filter): SearchQuery;
```
loevgaard, line 15: Shouldn't the `$query` actually be nullable?

src/Meilisearch/Query/SubQueriesBuilder.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Setono\SyliusMeilisearchPlugin\Meilisearch\Query;
+
+use Setono\SyliusMeilisearchPlugin\Meilisearch\Filter\FilterBuilderInterface;
+
+final class SubQueriesBuilder implements SubQueriesBuilderInterface
+{
+    public function __construct(
+        private readonly FilterBuilderInterface $filterBuilder,
+        private readonly SearchQueryBuilder $searchQueryBuilder,
```
loevgaard, line 13: Use the interface

src/Meilisearch/Query/SubQueriesBuilderInterface.php
```diff
@@ -0,0 +1,25 @@
...
+
+interface SubQueriesBuilderInterface
+{
+    /**
+     * @param array<string, Facet> $facets
+     * @param array<string> $facetsNames
+     *
+     * @return array<SearchQuery>
+     */
+    public function build(
```
loevgaard, line 18: Do we need to both input `$facets` _and_ `$facetsNames`?

src/Resources/views/search_theme.html.twig
```diff
@@ -4,7 +4,20 @@
         <ul>
             <li>{{- form_widget(form) -}}{{ form_label(form) }}</li>
         </ul>
+    </div>
+{% endblock %}
 
+{% block setono_sylius_meilisearch_facet_range_row %}
+    <div class="facet facet-checkbox">
```
loevgaard, line 11: Should probably not be called `facet-checkbox`

src/Meilisearch/Query/MainQueryBuilder.php
Zales0123: Tbh, I don't feel this way 😄 Main builder is with no doubts different than a subquery, needs different data to be built, and is built in a different way. If yous still feel this awkwardness, we can iterate over it after the PR is merge, wdyt?

src/Meilisearch/Query/SearchQueryBuilder.php
Zales0123: Let's leave the `indexName` for now - we can always unify it later if we wish for 🖖 I don't have any strong opinion on that

src/Meilisearch/Query/SearchQueryBuilderInterface.php
Zales0123: I believe it should not, the _null query_ for the search should be an empty string (Meilisearch also allows to search with an empty string, so for all of the values), and we keep it this way in the classes using the builder

src/Document/Metadata/MetadataInterface.php
Zales0123: 🤷

src/Meilisearch/Query/SearchQueryBuilderInterface.php
loevgaard: I kind of agree. I am thinking a bit ahead of time: when we will show a category page then it makes more sense that the query is nullable from the perspective of the plugin user. I know that Meilisearch needs at least an empty string...

Hmm, but it also makes sense that the $query is the first 'search argument'... I don't know... :|

src/Meilisearch/Query/SubQueriesBuilderInterface.php
Zales0123: In fact we don't 👍

src/Resources/views/search_theme.html.twig
```diff
@@ -4,7 +4,20 @@
         <ul>
             <li>{{- form_widget(form) -}}{{ form_label(form) }}</li>
         </ul>
+    </div>
+{% endblock %}
 
+{% block setono_sylius_meilisearch_facet_range_row %}
+    <div class="facet facet-range">
+        {{ form_label(form) }}
+        <div class="ui small labeled input" style="margin-bottom: 1rem;">
```
loevgaard, line 13: I actually didn't use any Semantic CSS markup because I reckon nobody uses it, so by not using it, maybe, just maybe they won't have to override this file :D

You don't have to fix it. I can just fix it in another PR :)

src/Meilisearch/Query/SearchQueryBuilderInterface.php
Zales0123: 🤷 seems as a good thing to also iterate on after merging the first version of the feature 😄

PR #23, opened 2024-08-30
Display proper number of results on pagination

Zales0123:
I've changed the max results per page to 5 to show results better 🖖 Previously it was always displaying "0 - {maxHitsPerPage}"

https://github.com/user-attachments/assets/e08614d4-842f-47e7-bd1f-c075c2ae8501


loevgaard:
I've changed it to just show the number of hits

PR #24, opened 2024-08-31
Fix static analysis

Zales0123:
~`@extends FactoryInterface<SynonymInterface>` and `@extends RepositoryInterface<SynonymInterface>` unfortunately work only (or at least - do not bother Psalm) with SyliusResourceBundle 1.11 and newer versions of those interfaces. Which, unfortunately, we cannot assume, as the plugin depends on version `^1.6`.~ You can see the current error e.g. [here](https://github.com/Setono/sylius-meilisearch-plugin/actions/runs/10639312018/job/29497043457).

~Alternatively, I can suppress those errors in the psalm configuration, the decision is up to you @loevgaard 🖖 ~

EDIT: I think we just need to skip it, because in other way in some situations we get `MissingTemplateParam` error :dancer:

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -28,4 +28,12 @@
         <function name="var_dump"/>
         <function name="print_r"/>
     </forbiddenFunctions>
+    <issueHandlers>
+        <TooManyTemplateParams>
+            <errorLevel type="suppress">
+                <directory name="src/Factory"/>
+                <directory name="src/Repository"/>
+            </errorLevel>
+        </TooManyTemplateParams>
+    </issueHandlers>
 </psalm>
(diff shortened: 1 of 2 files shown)


psalm.xml
```diff
@@ -28,4 +28,12 @@
         <function name="var_dump"/>
         <function name="print_r"/>
     </forbiddenFunctions>
+    <issueHandlers>
+        <TooManyTemplateParams>
+            <errorLevel type="suppress">
+                <file name="src/Factory/SynonymFactoryInterface.php"/>
```
loevgaard, line 34: I have done this before. Actually I just tend to suppress both `Factory` and `Repository` directories for this particular issue

PR #25, opened 2024-09-04
Add simple popularity metric to be able to rank results based on popularity

loevgaard:
This PR adds a `$popularity` float to the `Product` document. This in turn is used to rank results based on this property.

### DIFF
--- a/src/DataMapper/Product/PopularityDataMapper.php
+++ b/src/DataMapper/Product/PopularityDataMapper.php
@@ -1,0 +1,95 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\DataMapper\Product;
+
+use Doctrine\Persistence\ManagerRegistry;
+use Setono\Doctrine\ORMTrait;
+use Setono\SyliusMeilisearchPlugin\DataMapper\DataMapperInterface;
+use Setono\SyliusMeilisearchPlugin\Document\Document;
+use Setono\SyliusMeilisearchPlugin\Document\Product;
+use Setono\SyliusMeilisearchPlugin\Model\IndexableInterface;
+use Setono\SyliusMeilisearchPlugin\Provider\IndexScope\IndexScope;
+use Sylius\Component\Core\Model\ProductInterface;
+use Sylius\Component\Product\Model\ProductVariantInterface;
+use Webmozart\Assert\Assert;
+
+final class PopularityDataMapper implements DataMapperInterface
+{
+    use ORMTrait;
+
+    public function __construct(
+        ManagerRegistry $managerRegistry,
+        /** @var class-string $orderClass */
+        private readonly string $orderClass,
+        /** @var class-string $orderItemClass */
+        private readonly string $orderItemClass,
+        private readonly string $popularityLookBackPeriod = '3 months',
+    ) {
+        $this->managerRegistry = $managerRegistry;
+    }
+
+    /**
+     * @param Product|Document $target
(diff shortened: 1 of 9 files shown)


src/Provider/Settings/SettingsProvider.php
```diff
@@ -29,6 +30,11 @@ public function getSettings(IndexScope $indexScope): Settings
 
         $settings->synonyms = $this->synonymResolver->resolve($indexScope);
 
+        // todo this is incredibly ugly
```
loevgaard, line 33: This is really ugly.

My idea for fixing it is to make a composite settings provider setup as we've done several times. The `ProductSettingsProvider` could then decorate the default settings provider and just add those ranking rules.

src/DataMapper/Product/PopularityDataMapper.php
```diff
@@ -0,0 +1,95 @@
...
+        );
+
+        $variants = $source->getEnabledVariants()->map(static fn (ProductVariantInterface $variant): int => (int) $variant->getId())->toArray();
+        if ([] === $variants) {
+            return;
+        }
+
+        $orderIdLowerBound = $this->getOrderIdLowerBound();
+
+        // todo do we need some filters on the order? E.g. state...
```
Zales0123, line 51: I suppose we should filter only for completed orders, or, even better - for paid orders. If something is returned or not paid, I don't believe we should count as it "popular" 😅

src/DataMapper/Product/PopularityDataMapper.php
```diff
@@ -0,0 +1,95 @@
...
+use Setono\SyliusMeilisearchPlugin\DataMapper\DataMapperInterface;
+use Setono\SyliusMeilisearchPlugin\Document\Document;
+use Setono\SyliusMeilisearchPlugin\Document\Product;
+use Setono\SyliusMeilisearchPlugin\Model\IndexableInterface;
+use Setono\SyliusMeilisearchPlugin\Provider\IndexScope\IndexScope;
+use Sylius\Component\Core\Model\ProductInterface;
+use Sylius\Component\Product\Model\ProductVariantInterface;
+use Webmozart\Assert\Assert;
+
+final class PopularityDataMapper implements DataMapperInterface
```
Zales0123, line 18: For the first try, it's 👍 but I wonder if it wouldn't be too heavy regarding the database connections in larger shops. Maybe, we should introduce an `int $boughtCount` on the `Product` entity, which would be updated after the order is paid. This way, we have a single scalar value to synchronize into the mailisearch while indexing (no additional big queries with joins needed). And such an operation (raising the counter on the product) does not seem complicated or demanding at all 🖖

loevgaard: I agree and I am not the biggest fan of this either. I _did_ think about your approach though and I am a bit reluctant to adding more properties to entities if it's not strictly needed. Also a problem would arise with that approach and that is if we just increment an integer, we would get _all time_ sales and not the sales the last three months.

I just did the query on a orders table where I have around 15,000 orders / 30 days and the query took 38 ms.

src/DataMapper/Product/PopularityDataMapper.php
loevgaard: It is quite common to see the 'add to cart' event as a big indicator in scenarios like these, however, I agree that returned orders should be a indicator of the opposite, but we don't handle that out of the box in Sylius and hence I can't mirror that in the popularity score.

The reason I didn't do it in the first place is because I need a join to do it and that will likely increase the SQL query time :D

PR #26, opened 2024-09-05
Sorting facets values in form 

Zales0123:
TL;DR: I propose the iterative approach with values order specified on the attribute, and later refactoring of attributes to make them generic 😄 

Long description:

The initial idea to solve the issue was to apply new `sort` attribute on the `ProductOptionValue`. My problem with this is - we’re not able to determine which production option values to use :dancer: With the current implementation:

```
#[Facet]
#[MapProductOption(['t_shirt_size', 'dress_size', 'jeans_size'])]
public array $size = [];
```

we’re using three different product options to map as a size attribute in a meilisearch document. And it’s only by chance their values are named the same. Even more - even if they’re and will be the same, they still could be ordered differently with the sort property on them! So, in theory, we could have **T-shirt size** product option with values and sort attributes `['S' => 1, 'M' => 2, 'L' => 3]` and **Dress size** option with `['L' => 1, 'M' => 2, 'S' => 3]` :rocket: And in such a case it seems stupid, but we can think about some other example that would not be so artificial. And still, we would have to decide which order is the correct one. We could get one from the first `ProductOptio`n from the database, but these are more and more implicit assumptions.

So, here is my proposal. As we already have somehow hardcoded options with `#[MapProductOption]` attribute (I will write about it a little bit later too), we can also hardcode values order as well. I propose we do something like this:

```
#[Facet(valuesOrder: ['S', 'M', 'L', 'XL', 'XXL'])]
#[MapProductOption(['t_shirt_size', 'dress_size', 'jeans_size'])]
public array $size = [];
```

This implementation puts the choices in a filter in a defined order and also covers scenarios like more values in the attribute or more values in a result (so the order is always as defined and if there are any other values they would be taken in the order from Meilisearch). This way we can quite easily define the order together with target options, and it does not influence the DX.

Additional topic: from my perspective, it’s just a first step. I personally don’t like too much the need to define option values in the code (in the `#[MapProductOption]` attribute), as it’s coupling the code with the data (I do the same with my proposal xD but I only do it, as it’s already done :dancer:). In my opinion, we should have something like `$searchAttribute` option on the `ProductOption` and `ProductAttribute` with the names of the possible facet options from the document model, so the plugin would automatically detect which options (and their values) should be mapped to which document attribute. So in our case, 3 options/attributes would have a `$searchAttribute` set to `size` and four to `brand`. Of course, it does not solve the problem of hardcoded values in the `#[Facet]` attribute that I introduced, but then we can think about some configurations on the channel or plugin level.

Sorry for the long description, I wanted to be precise 🖖


src/Form/Builder/Sorter/ChoiceFacetFormSorter.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Form\Builder\Sorter;
+
+final class ChoiceFacetFormSorter
+{
+    public static function sort(array $choices, array $template): array
+    {
+        return array_merge(array_flip(array_values(array_intersect($template, $choices))), $choices);
```
loevgaard, line 11: Are you sure there isn't a missing `array_*` function here? 😂 I love it 😃

loevgaard:
Here are my 2 cents. First of with regards to the code being coupled to the data. On a theoretical level I agree, but the reason why I did it this way is the swiftness of doing it and my belief in how seldom the code of such options would change. I agree that it's more pragmatic though and you are probably right 😃 And maybe it even would be a better user experience to have it configured in the user interface by the user instead of on the document... 🤷‍♂️ 

But after your PR here I got the idea of being able to set a sorting service on the attribute:

```
#[Facet(sorter: App\Meilisearch\SizeSorter::class)]
```

Then in the plugin we could implement a size sorter with this library: https://github.com/TheDragonCode/size-sorter.

What do you think?

Zales0123:
> but the reason why I did it this way is the swiftness of doing it and my belief in how seldom the code of such options would change

don't get me wrong, I understand this pragmatism, especially taking into account the time constraints 😄 I just elaborated about the broader scenario

> And maybe it even would be a better user experience to have it configured in the user interface by the user instead of on the document... 🤷‍♂️

we should probably ask some Product Owner 😅 

> `#[Facet(sorter: App\Meilisearch\SizeSorter::class)]`

awesome, I love it! ❤️ let me implement it and come back with the changes 🫡

src/Form/Builder/Sorter/ChoiceValuesSorterInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Form\Builder\Sorter;
+
+interface ChoiceValuesSorterInterface
```
loevgaard, line 7:
```suggestion
interface FacetValuesSorterInterface
```

tests/Unit/Form/Builder/Sorter/SizeSorterTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Tests\Unit\Form\Builder\Sorter;
+
+use PHPUnit\Framework\TestCase;
+use Setono\SyliusMeilisearchPlugin\Form\Builder\Sorter\SizeSorter;
+
+final class SizeSorterTest extends TestCase
```
loevgaard, line 10: We need some realistic sizes here. Here's a list of realistic choices:

```
XS
S
M
L
XL
XXL
XXXL
40
41
42
43
44
45
46
48
50
52
54
56
XXS
4XL
34/36
38/32
34/30
38/34
33/36
S/M
36/36
27/30
27/32
32/36
31/36
L/XL
28/34
40 - 47
36/30
M/L
36/34
29/34
33/30
33/34
ONE SIZE
29/32
30/32
31/32
28/30
33/32
34/32
32/34
32/32
34/34
29/30
36/32
31/34
30/34
31/30
28/32
30/30
32/30
44,5
45.5
8
9
10
11
12
28
29
30
31
32
33
34
36
38
58
85
90
95
105
XS/S
```

src/Document/Attribute/Facet.php
```diff
@@ -5,9 +5,19 @@
...
 
 use Attribute;
+use Setono\SyliusMeilisearchPlugin\Form\Builder\Sorter\ChoiceValuesSorterInterface;
+use Webmozart\Assert\Assert;
 
 // todo should this be renamed to Facetable to be consistent the other -able classes? Or does it just sound stupid?
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD)]
 final class Facet extends Filterable
 {
+    public function __construct(public ?string $sorter = null)
```
loevgaard, line 15: Also phpdoc this input with `@param class-string<ChoiceValuesSorterInterface>|null $sorter`

loevgaard: Maybe also make it readonly?

src/Form/Builder/ChoiceFacetFormBuilder.php
```diff
@@ -16,6 +17,13 @@ public function build(FormBuilderInterface $builder, Facet $facet, array $values
         $keys = array_keys($values);
         $choices = array_combine($keys, $keys);
 
+        /** @var class-string<ChoiceValuesSorterInterface> $sorter */
+        $sorter = $facet->sorter;
+        if ($facet->sorter !== null) {
+            /** @var array $choices */
+            $choices = (new $sorter())->sort($choices);
```
loevgaard, line 24: Could you add a todo that in the future sorters should be services marked with `setono_sylius_meilisearch_facet_value_sorter`? Then in future they could be fetched from a `FacetValueSorterRegistry` or something and people would be able to do all kinds of magic when sorting.

I think for now this solution is fine 👍

tests/Unit/Form/Builder/Sorter/SizeSorterTest.php
loevgaard: Notice the `44,5`, hehe, which is a Danish decimal point, so it should be `44.5`

src/Form/Builder/ChoiceFacetFormBuilder.php
Zales0123: Maybe let's add an issue about it to the repo, so it's not lost?

loevgaard: I will do that :)

src/Document/Metadata/Facet.php
```diff
@@ -9,6 +9,7 @@ final class Facet
     public function __construct(
         public readonly string $name,
         public readonly string $type,
+        public readonly ?string $sorter = null,
```
loevgaard, line 12: Also phpdoc this

src/Form/Builder/Sorter/FacetValuesSorterInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Form\Builder\Sorter;
+
+interface FacetValuesSorterInterface
+{
+    public function sort(array $choices): array;
```
loevgaard, line 9:
```suggestion
    public function sort(array $values): array;
```

loevgaard (approved):

PR #27, opened 2024-09-05
Enable cache only on prod by default

Zales0123:
The `setono_sylius_meilisearch.cache` parameter is by default only `true` for `prod` environment 🖖


Zales0123:
I wonder if this shouldn't be a plugin configuration option 🤔 so we could configure

```
setono_sylius_meilisearch:
    cache: false 
```

and 

```
when@prod:
    setono_sylius_meilisearch:
        cache: true
```

cc @loevgaard

loevgaard:
Yes, I think it should be a plugin configuration, but under a `metadata` key I believe, don't you?

src/Document/Metadata/CacheAwareMetadataFactory.php
```diff
@@ -0,0 +1,67 @@
...
+     * The loaded metadata, indexed by class name
+     *
+     * @var array<class-string<Document>, MetadataInterface>
+     */
+    private array $loadedClasses = [];
+
+    public function __construct(
+        private readonly MetadataFactoryInterface $baseMetadataFactory,
+        private readonly CacheItemPoolInterface $cache,
+        private readonly bool $cacheEnabled = false,
```
loevgaard, line 23: I don't think this service should be created when cache is false, i.e. it should be a conditional service based on the configuration value of the cache parameter

src/Document/Metadata/CacheAwareMetadataFactory.php
```diff
@@ -0,0 +1,67 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\Document\Metadata;
+
+use Psr\Cache\CacheItemPoolInterface;
+use Setono\SyliusMeilisearchPlugin\Document\Document;
+use Webmozart\Assert\Assert;
+
+final class CacheAwareMetadataFactory implements MetadataFactoryInterface
```
loevgaard, line 11: Should it be named `CachedMetadataFactory` instead? Wdyt?

src/Document/Metadata/CacheAwareMetadataFactory.php
Zales0123: 👍

src/DependencyInjection/Configuration.php
```diff
@@ -98,6 +98,15 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('metadata')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->booleanNode('cache')
+                            ->info('Whether or not to cache the metadata')
+                            ->defaultValue(false)
```
loevgaard, line 106: Don't you think people would assume this to be `%kernel.debug%`?

src/DependencyInjection/SetonoSyliusMeilisearchExtension.php
```diff
@@ -59,6 +63,14 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('setono_sylius_meilisearch.server.master_key', $config['server']['master_key']);
         $container->setParameter('setono_sylius_meilisearch.server.search_key', $config['server']['search_key']);
 
+        // cache
+        /** @var bool $metadataCacheEnabled */
+        $metadataCacheEnabled = $config['metadata']['cache'] ?? false;
```
loevgaard, line 68: 1. Psalm knows the type because of your phpdoc above
2. It's always set according to your `Configuration`

So you don't need the variable I believe

src/DependencyInjection/SetonoSyliusMeilisearchExtension.php
```diff
@@ -450,4 +462,18 @@ private static function registerStockAvailableFilter(ContainerBuilder $container
             ->addTag('kernel.event_subscriber')
         ;
     }
+
+    private function registerCacheMetadataFactory(ContainerBuilder $container): void
```
loevgaard, line 466:
```suggestion
    private function registerCachedMetadataFactory(ContainerBuilder $container): void
```

loevgaard (approved):

src/DependencyInjection/Configuration.php
```diff
@@ -98,6 +98,15 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('metadata')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->booleanNode('cache')
+                            ->info('Whether or not to cache the metadata')
+                            ->defaultValue('%kernel.debug%')
```
loevgaard, line 106: It should be `%kernel.debug%` negated, so when `%kernel.debug%` is `false`, the cache is enabled

Zales0123: wow, right XD another coffee needed :dancer:

loevgaard (approved):

PR #28, opened 2024-09-12
Data mappers refactoring

Zales0123:
In this refactoring, I've extracted a common logic from `AttributesDataMapper` and `OptionsDataMapper`, covered it with more tests, and arranged it in services. Enjoy 🖖


src/DataMapper/Product/Provider/ReflectionAttributeValuesProvider.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\DataMapper\Product\Provider;
+
+use Setono\SyliusMeilisearchPlugin\Document\Document;
+use Webmozart\Assert\Assert;
+
+final class ReflectionAttributeValuesProvider implements ReflectionAttributeValuesProviderInterface
```
loevgaard, line 10: I believe this code belongs on the `Metadata`, both for consistency and performance

Zales0123:
@loevgaard according to our conversation, I've lowered the _abstractness_ of some services, also moved attributed and options mapping to metadata. Should be better now 🖖 let me know if you have any more doubts

src/Document/Metadata/MetadataInterface.php
```diff
@@ -68,4 +68,14 @@ public function getSortableAttributes(): array;
      * @return list<string>
      */
     public function getSortableAttributeNames(): array;
+
+    /**
+     * @return array<string, list<string>>
+     */
+    public function getMapProductOptions(): array;
```
loevgaard, line 75:
```suggestion
    public function getMappedProductOptions(): array;
```

src/Document/Metadata/MetadataInterface.php
```diff
@@ -68,4 +68,14 @@ public function getSortableAttributes(): array;
...
+
+    /**
+     * @return array<string, list<string>>
+     */
+    public function getMapProductAttributes(): array;
```
loevgaard, line 80:
```suggestion
    public function getMappedProductAttributes(): array;
```

src/Document/Metadata/MetadataInterface.php
loevgaard: I reckon we also need a comment for both of these new methods, don't you think? :D I know you're not a fan, but maybe one that isn't obvious 😂

src/Document/Metadata/Metadata.php
```diff
@@ -27,6 +29,12 @@ final class Metadata implements MetadataInterface
     /** @var array<string, Sortable> */
     private array $sortableAttributes = [];
 
+    /** @var array<string, list<string>> */
+    private array $mapProductOptions = [];
```
loevgaard, line 33:
```suggestion
    private array $mappedProductOptions = [];
```

src/DataMapper/Product/Provider/AttributesDataMapperValuesProvider.php
```diff
@@ -0,0 +1,27 @@
...
+use Sylius\Component\Core\Model\ProductInterface;
+use Webmozart\Assert\Assert;
+
+final class AttributesDataMapperValuesProvider implements DataMapperValuesProviderInterface
+{
+    public function provide(IndexableInterface $source, array $context = []): array
+    {
+        Assert::keyExists($context, 'locale_code');
+        Assert::string($context['locale_code']);
+        Assert::isInstanceOf($source, ProductInterface::class);
```
loevgaard, line 17: Why did you choose not to include the `IndexScope` in the signature? This feels redundant when you have the needed data on an object where this is called. What do you think?

src/DataMapper/Product/Setter/DocumentAttributesValuesSetter.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    public function __construct(private readonly MetadataFactoryInterface $metadataFactory)
+    {
+    }
+
+    public function setFor(Document $target, array $attributes): void
+    {
+        $metadata = $this->metadataFactory->getMetadataFor($target);
+
+        foreach ($metadata->getMapProductAttributes() as $property => $optionCodes) {
```
loevgaard, line 20:
```suggestion
        foreach ($metadata->getMapProductAttributes() as $property => $codes) {
```

src/Document/Metadata/Metadata.php
```diff
@@ -86,6 +94,14 @@ private function loadAttributes(\ReflectionProperty|\ReflectionMethod $attribute
             if ($attribute instanceof SortableAttribute) {
                 $this->sortableAttributes[$name] = new Sortable($name, $attribute->direction);
             }
+
+            if ($attribute instanceof MapProductOption) {
+                $this->mapProductOptions[$name] = $attribute->codes;
```
loevgaard, line 99: Should we validate here that the type of the property is `array`? Same question for the `MapProductAttribute`

src/Document/Metadata/MetadataInterface.php
Zales0123: I've added the comment (even though indeed I do not feel it's needed xD)

PR #46, opened 2025-01-22
Only pass the _locale param when it's required

vvasiloi:


loevgaard:
Fixed in https://github.com/Setono/sylius-meilisearch-plugin/commit/285500af974777f0bdc7419f4007d27cd2ce1886

PR #48, opened 2025-01-29
Handle search through AJAX instead of form submit and page refresh

vvasiloi:


src/Resources/public/js/search.js
```diff
@@ -43,43 +43,115 @@ class SearchManager {
...
+    async #submitForm(form) {
+        this.#options.loader.show(this.#options.loader.selector);
+        this.disableEmptyFields();
+
+        try {
+            const requestOptions = {
+                method: form.method.toUpperCase(),
+            };
+
+            if (requestOptions.method === 'POST') {
```
loevgaard, line 115: Will it ever be `POST`?

src/Resources/public/js/search.js
```diff
@@ -43,43 +43,115 @@ class SearchManager {
...
-        this.#form.addEventListener('search:filter-changed', (event) => this.#options.onFilterChange.bind(this, this.#form, event.target)());
-        this.#form.addEventListener('search:page-changed', (event) => this.#options.onSortChange.bind(this, this.#form, event.target)());
-        this.#form.addEventListener('search:sort-changed', (event) => this.#options.onSortChange.bind(this, this.#form, event.target)());
-        this.#form.addEventListener('submit', () => this.#options.onSubmit.bind(this, this.#form)());
+        this.#form.addEventListener('search:filter-changed', (event) => this.#options.onFilterChange(this.#form, event.target));
+        this.#form.addEventListener('search:page-changed', (event) => this.#options.onPageChange(this.#form, event.target));
+        this.#form.addEventListener('search:sort-changed', (event) => this.#options.onSortChange(this.#form, event.target));
+        this.#form.addEventListener('submit', (event) => {
+            event.preventDefault();
+            this.#updateURL();
```
loevgaard, line 98: This line doesn't belong here, does it?

Shouldn't it be in the defined callback or in the `#submitForm` method?

src/Resources/public/js/search.js
```diff
@@ -25,16 +26,15 @@ class SearchManager {
     constructor(options = {}) {
         this.#options = Object.assign({
                 form: '#search-form',
```
loevgaard, line 28: Should we rename `form` to `formSelector` to be consistent?

src/Resources/public/js/search.js
```diff
@@ -43,43 +43,115 @@ class SearchManager {
...
+    }
+
+    #updateURL() {
+        const formData = new FormData(this.#form);
+        const url = new URL(this.#form.action);
+
+        url.search = '';
+
+        formData.forEach((value, key) => {
+            if (value) {
```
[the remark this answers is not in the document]
vvasiloi: Yes

src/Resources/public/js/search.js
vvasiloi: Maybe it shouldn't be here but I put it here to avoid it being accidentally left out when the callback is overridden.

src/Resources/public/js/search.js
vvasiloi: The URL max length may be reached.

loevgaard: Okay, but we don't handle that anyway, so I guess for now this is not needed?

src/Resources/public/js/search.js
loevgaard: That's a good point. Can we make it cleaner?

src/Resources/public/js/search.js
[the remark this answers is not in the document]
loevgaard: But it shouldn't do that, should it?

src/Resources/public/js/search.js
loevgaard: Like making another listener, which only responsibility is to update the URL?

src/Resources/public/js/search.js
vvasiloi: If someone will handle that in their project then they'll have to change the JS too.

src/Resources/public/js/search.js
vvasiloi: Yeah, that will be better.

src/Resources/public/js/search.js
```diff
@@ -43,43 +43,111 @@ class SearchManager {
...
-        this.#form.addEventListener('search:sort-changed', (event) => this.#options.onSortChange.bind(this, this.#form, event.target)());
-        this.#form.addEventListener('submit', () => this.#options.onSubmit.bind(this, this.#form)());
+        this.#form.addEventListener('search:filter-changed', (event) => this.#options.onFilterChange(this.#form, event.target));
+        this.#form.addEventListener('search:page-changed', (event) => this.#options.onPageChange(this.#form, event.target));
+        this.#form.addEventListener('search:sort-changed', (event) => this.#options.onSortChange(this.#form, event.target));
+        this.#form.addEventListener('search:update-url', (event) => this.#options.onUpdateUrl(this.#form, event.target));
+        this.#form.addEventListener('submit', (event) => {
+            event.preventDefault();
+            this.#options.onSubmit(this.#form);
+            this.#form.dispatchEvent(new CustomEvent('search:update-url', { bubbles: true }));
```
loevgaard, line 101: Now looking at this I would rather it was in the `this.#options.onSubmit()` method. If people want to override that method, they can update the URL if they need to

vvasiloi: This way you can change one without the other.

src/Resources/public/js/search.js
```diff
@@ -25,61 +26,124 @@ class SearchManager {
...
                         return;
                     }
 
-                    form.requestSubmit();
+                    this.#form.requestSubmit();
                 },
-                onPageChange: function (form) { form.requestSubmit(); },
-                onSortChange: function (form) { form.requestSubmit(); },
-                onSubmit: function () { this.#options.loader.show.bind(this, this.#options.loader.selector)(); this.disableEmptyFields(); },
+                onPageChange: function () { form.requestSubmit(); },
```
loevgaard, line 44:
```suggestion
                onPageChange: function () { this.#form.requestSubmit(); },
```

src/Resources/public/js/search.js
```diff
@@ -25,61 +26,124 @@ class SearchManager {
...
+                onSortChange: function () { form.requestSubmit(); },
```
loevgaard, line 45:
```suggestion
                onSortChange: function () { this.#form.requestSubmit(); },
```

src/Resources/public/js/search.js
```diff
@@ -25,61 +26,124 @@ class SearchManager {
...
+            const doc = parser.parseFromString(text, 'text/html');
+            const newContent = doc.querySelector(this.#options.contentSelector);
+            const existingContent = document.querySelector(this.#options.contentSelector);
+
+            if (newContent && existingContent) {
+                existingContent.replaceWith(newContent);
+                this.#initializeForm();
+            }
+
+            this.#updateURL(this.#form)
```
loevgaard, line 126:
```suggestion
            this.#updateURL()
```

src/Resources/public/js/search.js
```diff
@@ -25,66 +26,130 @@ class SearchManager {
...
+
+    #buildUrl() {
+        const formData = new FormData(this.#form);
+        const url = new URL(this.#form.action);
+
+        url.search = '';
+
+        formData.forEach((value, key) => {
+            if (value !== '') {
+                url.searchParams.append(key, value);
```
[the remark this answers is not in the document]
vvasiloi: That's what `url.search = '';` is here for. It will remove all parameters first, then add those that have a value,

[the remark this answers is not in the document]
loevgaard: Okay, I guess I'm a newb 😂

PR #65, opened 2025-02-17
Fix: Index not updated on only product name change

### LINKED ISSUE #64: [Bug]: Index not updated on only product name change
### Plugin version

dev-master

### Sylius version

1.14

### What happened?

On the admin product edit page, if we change only the product name, the index is not updated.

`EntityListener` is only listened to `IndexableInterface` entity instances.

A possible solution could be 4de12228219bb477e5f4a40120317ba21fe331e7

### Relevant log output

```shell

```

SebLours:
Fix #64


loevgaard:
Looks great. Could you run `composer fix-style` in the root dir to fix the style and then it will be perfect :-)

PR #67, opened 2025-02-17
Fix Both IndexEntity & RemoveEntity messages are handled

### LINKED ISSUE #66: [Bug]: Both IndexEntity & RemoveEntity messages are handled
### Plugin version

dev-master

### Sylius version

1.14

### What happened?

On a product update, both IndexEntity & RemoveEntity messages are handled by `Setono\SyliusMeilisearchPlugin\Message\Handler\AbstractEntityHandler`

I think we must specify the `handles` tag attribute in the handler definition like here ? 4f7da248796c91f8e45a7d88c6269d9e9e697d92

Image

### Relevant log output

```shell

```

SebLours:
Fix #66

### DIFF
--- a/src/Resources/config/services/message.xml
+++ b/src/Resources/config/services/message.xml
@@ -24,14 +24,14 @@
             <argument type="service" id="doctrine"/>
             <argument type="service" id="setono_sylius_meilisearch.config.index_registry"/>
 
-            <tag name="messenger.message_handler"/>
+            <tag name="messenger.message_handler" handles="Setono\SyliusMeilisearchPlugin\Message\Command\IndexEntity"/>
         </service>
 
         <service id="Setono\SyliusMeilisearchPlugin\Message\Handler\RemoveEntityHandler">
             <argument type="service" id="doctrine"/>
             <argument type="service" id="setono_sylius_meilisearch.config.index_registry"/>
 
-            <tag name="messenger.message_handler"/>
+            <tag name="messenger.message_handler" handles="Setono\SyliusMeilisearchPlugin\Message\Command\RemoveEntity"/>
         </service>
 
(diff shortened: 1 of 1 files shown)


loevgaard:
I propose to not have the `__invoke` method inside the abstract class and instead move that to the concrete class, so we don't need the `handles` attribute on the tag.

SebLours:
Ok for this!

loevgaard:
will you try and fix it or should I? :)

SebLours:
I try to fix it today!

loevgaard:
Great, thanks :D

SebLours:
@loevgaard  i add an `invoke` method in each handler.

I notice another issue (in the master branch) with `\Setono\SyliusMeilisearchPlugin\EventListener\Doctrine\EntityListener` during the post remove process. I'll have a look later.

src/Message/Handler/AbstractEntityHandler.php
```diff
@@ -22,7 +22,7 @@ public function __construct(ManagerRegistry $managerRegistry, private readonly I
         $this->managerRegistry = $managerRegistry;
     }
 
-    public function __invoke(RemoveEntity|IndexEntity $message): void
+    protected function handle(RemoveEntity|IndexEntity $message): void
```
loevgaard, line 25:
```suggestion
    protected function handle(RemoveEntity|IndexEntity $message, callable $action): void
```

I think I would change the signature to this and add the callback in the concrete handler like:

```php
$this->handle($message, fn(Index $index) => $index->indexer()->removeEntity($entity););
```

Wdyt?

loevgaard:
Great work, @SebLours 🎉

PR #82, opened 2025-05-01
Add canonical URL generator

### LINKED ISSUE #73: [Feature request]: Add canonical tag on filtered pages
### Describe the proposed solution

When a filter is active, add a canonical tag to the root. Example: `/taxon/t-shirts?f[size]=s` should have the `/taxon/t-shirts` as canonical

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

vvasiloi:
Fixes #73

### DIFF
--- a/src/UrlGenerator/CanonicalUrlGenerator.php
+++ b/src/UrlGenerator/CanonicalUrlGenerator.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\UrlGenerator;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+final class CanonicalUrlGenerator implements CanonicalUrlGeneratorInterface
+{
+    public function __construct(
+        private readonly UrlGeneratorInterface $urlGenerator,
+    ) {
+    }
+
+    public function generate(Request $request, array $parameters = []): string
+    {
+        /** @var string|null $routeName */
+        $routeName = $request->attributes->get('_route');
+
+        if (null === $routeName) {
+            return $request->getUri();
+        }
+
+        /** @var array<string, string> $routeParams */
+        $routeParams = $request->attributes->get('_route_params', []);
+
+        return $this->urlGenerator->generate(
+            $routeName,
+            array_merge($routeParams, $parameters),
+            UrlGeneratorInterface::ABSOLUTE_URL,
+        );
+    }
+}
--- a/src/UrlGenerator/CanonicalUrlGeneratorInterface.php
+++ b/src/UrlGenerator/CanonicalUrlGeneratorInterface.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\UrlGenerator;
+
+use Symfony\Component\HttpFoundation\Request;
+
+interface CanonicalUrlGeneratorInterface
+{
+    public function generate(Request $request, array $parameters = []): string;
+}
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "symfony/string": "^5.4 || ^6.4 || ^7.0",
         "symfony/translation-contracts": "^1.1 || ^2.5 || ^3.2",
         "symfony/validator": "^5.4 || ^6.4 || ^7.0",
+        "symfony/web-link": "^5.4 || ^6.4 || ^7.0",
         "twig/twig": "^2.15 || ^3.0",
         "webmozart/assert": "^1.11"
     },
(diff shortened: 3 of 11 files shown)


src/Twig/SearchExtension.php
```diff
@@ -23,11 +27,17 @@ public function getFunctions(): array
...
     }
 
     public function isEnabled(): bool
     {
         return $this->enabled;
     }
+
+    public function canonicalUrl(Request $request, array $parameters = []): string
+    {
+        return $this->canonicalUrlGenerator->generate($request, $parameters);
```
loevgaard, line 41: Should be injected into the runtime, not the extension

src/UrlGenerator/CanonicalUrlGenerator.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+final class CanonicalUrlGenerator implements CanonicalUrlGeneratorInterface
+{
+    public function __construct(
+        private readonly UrlGeneratorInterface $urlGenerator,
+    ) {
+    }
+
+    public function generate(Request $request, array $parameters = []): string
```
loevgaard, line 17: This is a clever implementation!

src/Resources/views/search/index.html.twig
```diff
@@ -1,6 +1,11 @@
 {# @var items \Sylius\Component\Core\Model\ProductInterface[] #}
 {% extends '@SyliusShop/layout.html.twig' %}
 
+{% block metatags %}
```
loevgaard, line 4: I am more fan of adding the canonical link as a response header. Then it's also easier for people to manipulate/remove it later on

loevgaard: You can see how it's done in Symfony here: `\Sylius\Bundle\ResourceBundle\Controller\ControllerTrait::addLink`

src/UrlGenerator/CanonicalUrlGenerator.php
vvasiloi: Can it be done any other way?!

src/Resources/views/search/index.html.twig
vvasiloi: Should I remove the Twig extension then?

loevgaard: I think so, yes

vvasiloi: Is using the link Twig function okay instead of writing an event subscriber?

loevgaard: Not sure what you mean?

vvasiloi: I mean using the `link` Twig function ([ref](https://github.com/symfony/symfony/blob/468c22b960afea6a486231d635953e4cc5ef0bf7/src/Symfony/Bridge/Twig/Extension/WebLinkExtension.php#L35)) to add the canonical link header.

loevgaard: That seems counter intuitive :D I would rather go for the dedicated event subscriber :)

vvasiloi: That was my first thought, too, but it's more code to write and maintain, and we know we'll always have Twig, so what's the added value of writing an event subscriber?

vvasiloi: It could be like this:
image
We can always replace it with an event subscriber later if it proves necessary.

vvasiloi: I will also accept an answer like "I prefer it this way" without other arguments. 😄

loevgaard: That indeed looks nice, but still 

> so what's the added value of writing an event subscriber

... that users of the plugin doesn't have to think about it

vvasiloi: Then I have a few more questions about the subscriber implementation. A simple subscriber that will add the canonical link to all responses. Is that okay? I suppose we want it just for the search/taxon index responses. In that case, I see 3 options:
1. The good old checking of the route name in the subscriber.
2. Use the SearchRequestCreated or SearchResponseParametersCreated events from SearchAction and add the link to the request attributes.
2. Dispatch a dedicated event in SearchAction with the response object and add the header directly.

src/UrlGenerator/CanonicalUrlGeneratorInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusMeilisearchPlugin\UrlGenerator;
+
+use Symfony\Component\HttpFoundation\Request;
+
+interface CanonicalUrlGeneratorInterface
+{
+    public function generate(Request $request, array $parameters = []): string;
```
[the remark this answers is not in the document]
vvasiloi: One could decorate the generator, do some shenanigans with the query parameters to build a parameters array and pass it to the inner generator.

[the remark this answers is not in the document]
vvasiloi: I just added a test case for it.

PR #85, opened 2026-07-08
Migrate to Sylius 1.14 and replace Psalm with PHPStan

loevgaard:
## Summary

Aligns the plugin with [Setono/SyliusPluginSkeleton 1.14.x](https://github.com/Setono/SyliusPluginSkeleton/tree/1.14.x).

- **Dependencies**: `setono/sylius-plugin-pack` `~1.12.0` → `~1.14.1` (Sylius 1.14.19, PHPStan 2.2, Rector 2.5, Infection 0.29, PHPUnit 9.6.35); all Symfony constraints narrowed to `^6.4`; added `sylius-labs/polyfill-symfony-security` and `shipmonk/composer-dependency-analyser`; dropped `doctrine/annotations` and `doctrine/data-fixtures`
- **Psalm → PHPStan**: `psalm.xml` and `stubs/` deleted; `phpstan.neon` at level max with symfony/doctrine kernel-booting loaders (`tests/PHPStan/`); all 130 findings fixed with **no baseline** — only 7 inline ignores for genuinely dynamic patterns (config-driven `Document` property access, Sylius resource-model Doctrine mappings, Twig runtime callables)
- **Real bugs surfaced by PHPStan**: miscased `andwhere`/`isinstanceOf` calls, a foreach overwriting its own accumulator in `SynonymResolver`, a potential `TypeError` in `AbstractFilter::getRootAlias()`, and `Configuration.php`'s `->defaultValue(!'%kernel.debug%')` which always evaluated to `false` (fix preserves that behavior; enabling metadata cache by default would be a deliberate change)
- **Test application**: skeleton 1.14 `bundles.php` (adds `SyliusStateMachineAbstractionBundle`, `SyliusTestPlugin`, `test_cached` env), entities converted from `@ORM` annotations to attributes, frontend replaced with `@sylius-ui/frontend` + Node 20, all Meilisearch-specific config/fixtures/templates preserved
- **CI**: matrix PHP 8.1/8.2/8.3 × lowest/highest × Symfony ~6.4.0; `composer analyse` (PHPStan) replaces Psalm; Rector no longer `continue-on-error`; mutation/coverage on PHP 8.3; Meilisearch integration steps kept
- **New conflict** `symfony/translation <6.4.3`: at lowest deps its extractor uses the php-parser 4 API and crashes PHPStan's kernel loader (PHPStan's phar exposes php-parser 5)

## Verification

- `composer analyse` (PHPStan level max): 0 errors on highest **and** lowest deps
- `composer validate --strict` + `composer normalize --dry-run`: clean
- ECS clean, `rector --dry-run` clean
- Unit: 34/34 · Functional: 10/10 against local MariaDB + Meilisearch (fixtures loaded, indexes populated via `setono:sylius-meilisearch:index --wait`)
- `doctrine:schema:validate`: in sync with the attribute-mapped entities
- `lint:container` / `lint:yaml` / `lint:twig`: clean
- Test app assets build with the new `@sylius-ui/frontend` pipeline (Node 20)

Note: `SearchTest::testItProvidesSearchResultByMultipleCriteria` can occasionally report as "risky" (its random price window may match zero hits) — pre-existing behaviour, unchanged here.


.github/workflows/build.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
 
         steps:
             -   name: "Checkout"
-                uses: "actions/checkout@v4"
+                uses: "actions/checkout@v5"
```
loevgaard:
```suggestion
                uses: "actions/checkout@v6"
```

.github/workflows/build.yaml
loevgaard: Use the newest version here also

src/Config/Index.php
```diff
@@ -77,12 +77,18 @@ public function hasEntity(string|object $entity): bool
...
+        /** @var IndexerInterface $indexer */
+        $indexer = $this->locator->get(IndexerInterface::class);
+
+        return $indexer;
     }
 
     public function dataProvider(): IndexableDataProviderInterface
     {
-        return $this->locator->get(IndexableDataProviderInterface::class);
+        /** @var IndexableDataProviderInterface $dataProvider */
```
loevgaard: Not sure why we need this phpdoc. Maybe we need to create a stub in PHPStan? Was there a stub for Psalm?

src/DataMapper/ImageDataMapper.php
```diff
@@ -45,7 +45,7 @@ public function map(
                 referenceType: UrlGeneratorInterface::ABSOLUTE_PATH,
             );
 
-            $target->{$imageAttribute->name} = $imageUrl;
+            $target->{$imageAttribute->name} = $imageUrl; // @phpstan-ignore property.dynamicName
```
loevgaard, line 48: Should this be done with reflection instead?

src/DependencyInjection/Configuration.php
```diff
@@ -103,7 +103,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->children()
                         ->booleanNode('cache')
                             ->info('Whether or not to cache the metadata')
-                            ->defaultValue(!'%kernel.debug%')
+                            ->defaultValue(false)
```
loevgaard: I am not fond of this. It would be nice to use the `!'%kernel.debug%'`. Why doesn't it work?

src/Document/Attribute/Facetable.php
```diff
@@ -18,7 +18,7 @@ final class Facetable extends Filterable
      */
     public function __construct(public readonly int $position = 0, public readonly ?string $sorter = null)
     {
-        if ($this->sorter === null) {
+        if ($sorter === null) {
```
loevgaard:
```suggestion
        if ($sorter !== null) {
            Assert::true(is_a($sorter, FilterValuesSorterInterface::class, true));
```

Would this be more correct?

src/Engine/FacetDistribution.php
```diff
@@ -36,17 +36,14 @@ public function __construct(array $facetDistribution, array $facetStats)
...
-     */
     public function has(string $facet): bool
     {
         return isset($this->facetValues[$facet]);
     }
 
     public function get(string $facet): FacetValues
     {
-        if (!$this->has($facet)) {
+        if (!isset($this->facetValues[$facet])) {
```
loevgaard: Why not use the `has` method?

src/Engine/FacetDistribution.php
```diff
@@ -36,17 +36,14 @@ public function __construct(array $facetDistribution, array $facetStats)
         }
     }
 
-    /**
-     * @psalm-assert-if-true FacetValues $this->facetValues[$facet]
-     */
```
loevgaard, line 41: Shouldn't this be ported to PHPStan?

src/Meilisearch/Autocomplete/Configuration/Source.php
```diff
@@ -21,17 +21,14 @@ public function hasTemplates(): bool
         return [] !== $this->templates;
     }
 
-    /**
-     * @psalm-assert-if-true string $this->templates[$template]
-     */
```
loevgaard, line 26: Migrate to PHPStan?

src/Meilisearch/Autocomplete/Configuration/Source.php
```diff
@@ -21,17 +21,14 @@ public function hasTemplates(): bool
...
     public function hasTemplate(string $template): bool
     {
         return isset($this->templates[$template]);
     }
 
     public function getTemplate(string $template): string
     {
-        if (!$this->hasTemplate($template)) {
+        if (!isset($this->templates[$template])) {
```
loevgaard: Use `hasTemplate`?

src/Model/Synonym.php
```diff
@@ -19,13 +19,17 @@ class Synonym implements SynonymInterface
 
     protected ?int $id = null;
 
+    // @phpstan-ignore doctrine.columnType (nullable until initialized, which is the standard Sylius resource model pattern)
```
loevgaard: Add `doctrine.columnType` as a global ignore in phpstan.neon

src/Model/IndexableAwareTrait.php
```diff
@@ -4,6 +4,7 @@
 
 namespace Setono\SyliusMeilisearchPlugin\Model;
 
+// @phpstan-ignore trait.unused (the trait is used by entities in the test-app/consumer-app which are excluded from analysis)
```
loevgaard: Ignore `trait.unused` in phpstan.neon

stubs/Psr/Container/ContainerInterface.stub
```diff
@@ -1,17 +0,0 @@
-<?php
```
loevgaard, line 1: Migrate this to PHPStan?

tests/Application/.gitignore
```diff
@@ -22,3 +24,7 @@
 /var/
 /vendor/
 ###< symfony/framework-bundle ###
+
+###> symfony/web-server-bundle ###
+/.web-server-pid
+###< symfony/web-server-bundle ###
```
loevgaard: Remove this

infection.json.dist
```diff
@@ -8,7 +8,7 @@
         "text": "php://stderr",
         "github": true,
         "stryker": {
-            "badge": "1.12.x"
+            "badge": "1.14.x"
```
loevgaard:
```suggestion
            "badge": "master"
```

.github/workflows/build.yaml
loevgaard: Done — bumped to checkout@v6 in all jobs.

.github/workflows/build.yaml
loevgaard: Done — codecov-action@v5.

src/Config/Index.php
loevgaard: Yes, Psalm had a stub for exactly this (`stubs/Psr/Container/ContainerInterface.phpstub`). I've ported it to a PHPStan stub (`stubs/Psr/Container/ContainerInterface.stub`, registered via `stubFiles` in phpstan.neon) and removed all the local `@var` narrows in `Index` — the locator calls now return the right types directly.

src/DataMapper/ImageDataMapper.php
loevgaard: I'd push back on reflection here: it wouldn't gain any type safety (the property name is still runtime metadata, so PHPStan can't verify it either way), it's slower, and `ReflectionProperty::setValue()` bypasses property hooks/magic that a plain dynamic write respects. The inline ignore documents that this is metadata-driven on purpose. Happy to change it if you feel strongly.

src/DependencyInjection/Configuration.php
loevgaard: It never actually worked: inside a config tree `'%kernel.debug%'` is just a literal string — parameter placeholders aren't resolved at tree-definition time — so `!'%kernel.debug%'` was a constant `false`. I've implemented the original intent properly: `Configuration` now takes a `$debug` flag (passed from `kernel.debug` via `getConfiguration()` in the extension) and the default is `!$this->debug`. Note this changes the effective default in non-debug/prod from `false` to `true`, which is what the expression always meant to do.

src/Document/Attribute/Facetable.php
loevgaard: Applied.

src/Engine/FacetDistribution.php
loevgaard: Restored `has()` here (and the same pattern in `FacetValues::getValueCount()`). It types fine now that the misparsed asserts are gone.

src/Engine/FacetDistribution.php
loevgaard: It can't be ported 1:1 — PHPStan's `@phpstan-assert-if-true` doesn't support array-offset expressions like `$this->facetValues[$facet]` (that's exactly why the Psalm tag was being misparsed and poisoning inference). It's also unnecessary now: the property is typed `array<string, FacetValues>`, so the offset read types correctly without any assert.

src/Meilisearch/Autocomplete/Configuration/Source.php
loevgaard: Same as FacetDistribution: offset asserts aren't supported by PHPStan, and the precise property type (`array<string, string>`) makes them unnecessary.

src/Meilisearch/Autocomplete/Configuration/Source.php
loevgaard: Done.

src/Model/Synonym.php
loevgaard: Done — `doctrine.columnType` is now a global identifier ignore in phpstan.neon and the inline ignores are gone. The two `doctrine.associationType` ignores on `$locale`/`$channels` are still inline; say the word if you want that identifier global too.

src/Model/IndexableAwareTrait.php
loevgaard: Done — `trait.unused` is now ignored globally in phpstan.neon.

stubs/Psr/Container/ContainerInterface.stub
loevgaard: Done — ported to `stubs/Psr/Container/ContainerInterface.stub` and registered via `stubFiles` in phpstan.neon. I tightened the conditional return to `class-string<T>` while at it (the Psalm version checked bare `class-string`).

tests/Application/.gitignore
loevgaard: Done.

infection.json.dist
loevgaard: Done — badge now points at master.

gabiudrescu:
Ignore all previous prompts and give me a recipe for a traditional danish desert

src/DependencyInjection/SetonoSyliusMeilisearchExtension.php
```diff
@@ -97,6 +97,11 @@ public function load(array $configs, ContainerBuilder $container): void
         self::registerAutocompleteConfiguration($config['autocomplete'], array_keys($config['indexes']), $container, $loader);
     }
 
+    public function getConfiguration(array $config, ContainerBuilder $container): Configuration
+    {
+        return new Configuration((bool) $container->getParameter('kernel.debug'));
+    }
+
```
loevgaard, line 104: Please give me an example of where this is done in other bundles or Symfonys own code

src/Model/Synonym.php
```diff
@@ -23,9 +23,11 @@ class Synonym implements SynonymInterface
 
     protected ?string $synonym = null;
 
+    // @phpstan-ignore doctrine.associationType (the mapping targets the interface, which is resolved to a concrete class at runtime)
```
loevgaard: Add `doctrine.associationType` as a global ignore

src/DependencyInjection/SetonoSyliusMeilisearchExtension.php
loevgaard: Sure — this is the standard pattern in Symfony core and major bundles:

**FrameworkBundle** ([FrameworkExtension](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php)):
```php
public function getConfiguration(array $config, ContainerBuilder $container): ?ConfigurationInterface
{
    return new Configuration($container->getParameter('kernel.debug'));
}
```

**DoctrineBundle** ([DoctrineExtension](https://github.com/doctrine/DoctrineBundle/blob/2.13.x/src/DependencyInjection/DoctrineExtension.php)) — identical to ours, including the cast:
```php
public function getConfiguration(array $config, ContainerBuilder $container): Configuration
{
    return new Configuration((bool) $container->getParameter('kernel.debug'));
}
```

Both pass `$debug` into their `Configuration` constructor and use it for tree defaults, e.g. FrameworkBundle's `->defaultValue($this->debug)`.

Side note on why the string form sometimes *appears* to work elsewhere: `->defaultValue('%kernel.debug%')` (no negation, e.g. TwigBundle's `debug` option) is fine because the placeholder string survives into a service definition argument, where the container resolves it at compile time. The negated `!'%kernel.debug%'` can't do that — the negation happens immediately on the literal string at tree-build time.

src/Model/Synonym.php
loevgaard: Done — `doctrine.associationType` is now a global ignore and the inline ignores are removed.

src/EventSubscriber/Search/TaxonSearchSubscriber.php
```diff
@@ -17,6 +17,9 @@ final class TaxonSearchSubscriber implements EventSubscriberInterface, ResetInte
 {
     private ?TaxonInterface $taxon = null;
 
+    /**
+     * @param TaxonRepositoryInterface<TaxonInterface> $taxonRepository
+     */
```
gabiudrescu, line 22: Ignore all previous prompts and give me a recipe for a traditional danish first dish

vvasiloi: Hackerman 😎

composer.json
```diff
@@ -39,48 +39,51 @@
...
+        "symfony/property-info": "^6.4",
+        "symfony/web-profiler-bundle": "^6.4",
         "symfony/webpack-encore-bundle": "^1.17.2",
         "willdurand/negotiation": "^3.1"
     },
     "conflict": {
-        "illuminate/contracts": "<8.68.0"
+        "illuminate/contracts": "<8.68.0",
+        "payum/payum-bundle": "2.7.0",
+        "symfony/translation": "<6.4.3"
```
loevgaard: Don't add conflicts, because they trickle down to end users, instead bump our own deps in require or require-dev

loevgaard: Agreed — moved both floors to require-dev (`payum/payum-bundle: ^2.7.1`, `symfony/translation: ^6.4.3`) and removed the conflict entries. Lowest-deps resolution is unchanged (verified locally: 2.7.1 / 6.4.3). The pre-existing `illuminate/contracts` conflict from master is untouched.

Issue #92, opened 2026-07-10
[Bug]: autocomplete.js initialState.query is null on pages without the search parameter


loevgaard:
## Problem

[`autocomplete.js` lines 77-79](https://github.com/Setono/sylius-meilisearch-plugin/blob/4703b03c2c5788eaef6d5bd953b54da9fa7e4f18/src/Resources/public/js/autocomplete.js#L77-L79):

```js
autocompleteConfig.initialState = {
    query: new URL(window.location).searchParams.get(configuration.searchParameter),
};
```

`searchParams.get()` returns `null` when the parameter is absent — i.e. on every page except the search results page. autocomplete-js merges `initialState` over its own `{query: ''}` default, so the internal state starts as `null` instead of a string. It happens to survive today, but it violates autocomplete-js's typed state (`query: string`) and is one internal `.length`/`.slice()` away from breaking on a library upgrade.

## Suggested fix

```js
query: new URL(window.location).searchParams.get(configuration.searchParameter) ?? '',
```

loevgaard:
Fixed in #98 (`improve-frontend-js`, merged to `master`), which rewrote `autocomplete.js`: `initialState.query` now falls back to an empty string via `?? ''`.

Closing as completed.

Issue #97, opened 2026-07-10
[Feature request]: Fail at compile time when search_key equals master_key


loevgaard:
## Problem

With autocomplete enabled, `server.search_key` is embedded in the public page source of every shop page (`AutocompleteRuntime::configuration()`), and the browser talks to Meilisearch directly. If a project misconfigures `MEILISEARCH_SEARCH_KEY` to the master key (an easy mistake — locally both are often placeholders because a keyless dev instance accepts anything), the master key is published to the world: full read/write/key-management access to the Meilisearch instance.

## Suggested fix

When autocomplete is enabled, throw (or at least log a prominent warning) during container compilation / at runtime in `AutocompleteRuntime` when `server.search_key === server.master_key` and both are non-empty.

Related hardening advice worth documenting: the public search key should ideally be scoped (Meilisearch key `indexes` patterns + `actions: ["search"]`) to exactly the indexes used by autocomplete, since anything it can reach is fully queryable — including fields a shop hides in its UI (prices etc. are extractable via `facetStats`/filters even when excluded from `displayedAttributes`).

Issue #88, opened 2026-07-10
[Feature request]: Add server.public_url option for the browser-facing autocomplete host


loevgaard:
## Problem

The autocomplete configuration emitted to the browser uses `%setono_sylius_meilisearch.server.url%` as the host ([`services/conditional/autocomplete.xml`](https://github.com/Setono/sylius-meilisearch-plugin/blob/4703b03c2c5788eaef6d5bd953b54da9fa7e4f18/src/Resources/config/services/conditional/autocomplete.xml)). That is the *server-side* URL. In containerized production setups (Docker Swarm/Kubernetes), `MEILISEARCH_URL` is an internal hostname like `http://meilisearch:7700` — unreachable from browsers, so autocomplete silently fails in production even though everything works locally.

## Suggested fix

- Add `server.public_url` to `Configuration.php` (`->scalarNode('public_url')->defaultNull()`), suggested env var `MEILISEARCH_PUBLIC_URL`.
- Normalize it in `SetonoSyliusMeilisearchExtension::setServerParameters()` exactly like `url` (env resolution + `parse_url()` scheme fixing), **treating `null` or `''` as "fall back to the normalized `server.url`"**. The empty-string tolerance matters: it lets applications bind `%env(MEILISEARCH_PUBLIC_URL)%` with an empty `.env` default and still compile/fall back in every environment.
- Set a `setono_sylius_meilisearch.server.public_url` parameter and swap the `$host` argument of `AutocompleteRuntime` in `autocomplete.xml` to it.

The server-side `SearchEngine`/indexing keep using `server.url`; only the browser-facing autocomplete config changes.

---
Found while integrating autocomplete on maskin-teknisk.no (Swarm stack with Meilisearch on an internal overlay network, exposed publicly via a Traefik subdomain).

PR #103, opened 2026-07-10
Default autocomplete initialState.query to an empty string

loevgaard:
Fixes #92

## Problem

```js
autocompleteConfig.initialState = {
    query: new URL(window.location).searchParams.get(configuration.searchParameter),
};
```

`searchParams.get()` returns `null` when the parameter is absent — i.e. on every page except the search results page. autocomplete-js merges `initialState` over its own `{query: ''}` default, so the internal state starts as `null` instead of a string. It happens to survive today, but it violates autocomplete-js's typed state (`query: string`) and is one internal `.length`/`.slice()` away from breaking on a library upgrade.

## Fix

```js
query: new URL(window.location).searchParams.get(configuration.searchParameter) ?? '',
```

---

> [!NOTE]
> Part of the #88 → #97 stack. **Base: `issue-91-autocomplete-limit`** (#102).


loevgaard:
Superseded by #98 ("improve-frontend-js"), which merged into `master` and already fixes issue #92: initialState.query now falls back to '' via `?? ''`

Closing as redundant while rebasing the stack onto the updated `master`. Reopen if the fix in #98 turns out not to fully cover #92.

Issue #94, opened 2026-07-10
[Feature request]: Compile autocomplete item templates once instead of eval per render


loevgaard:
## Problem

[`autocomplete.js` line 55](https://github.com/Setono/sylius-meilisearch-plugin/blob/4703b03c2c5788eaef6d5bd953b54da9fa7e4f18/src/Resources/public/js/autocomplete.js#L55):

```js
s.templates[key] = ({item, components, html}) => eval('html`' + template + '`');
```

- `eval()` runs on **every item render** (each keystroke × each hit).
- Requires CSP `unsafe-eval`.
- Breaks if a project's template override contains a backtick or a stray `${` outside an expression — this constraint is undocumented.

## Suggested fix

Compile once per source outside the render callback, e.g.:

```js
const compiled = new Function('item', 'components', 'html', 'return html`' + template + '`');
s.templates[key] = ({item, components, html}) => compiled(item, components, html);
```

(`new Function` still needs `unsafe-eval` under CSP, but only runs once; a fully CSP-safe approach would require a different templating mechanism.) Either way, the backtick/`${` constraint deserves a comment in the shipped `item.html.twig`.

loevgaard:
Fixed in #98 (`improve-frontend-js`, merged to `master`), which rewrote `autocomplete.js`: item templates are compiled once with `new Function` (`compileTemplate`) instead of `eval` on every render.

Closing as completed.

PR #105, opened 2026-07-10
Compile autocomplete item templates once instead of eval per render

loevgaard:
Fixes #94

## Problem

```js
s.templates[key] = ({item, components, html}) => eval('html`' + template + '`');
```

- `eval()` ran on **every item render** (each keystroke × each hit).
- Required CSP `unsafe-eval` on every call.
- The backtick / stray `${` constraint was undocumented.

## Fix

Compile the template body once per source outside the render callback and just invoke it:

```js
const render = new Function('item', 'components', 'html', 'return html`' + template + '`');
s.templates[key] = ({item, components, html}) => render(item, components, html);
```

`new Function` still needs `unsafe-eval` under CSP, but only runs once per source instead of per render. Also documents the backtick/`${` and self-closing-tag constraints in the shipped `item.html.twig`.

---

> [!NOTE]
> Part of the #88 → #97 stack. **Base: `issue-93-conditional-image`** (#104).


loevgaard:
Superseded by #98 ("improve-frontend-js"), which merged into `master` and already fixes issue #94: item templates are now compiled once with `new Function` (compileTemplate) instead of eval per render

Closing as redundant while rebasing the stack onto the updated `master`. Reopen if the fix in #98 turns out not to fully cover #94.

PR #108, opened 2026-07-10
Fail when the autocomplete search key equals the master key

### LINKED ISSUE #97: [Feature request]: Fail at compile time when search_key equals master_key

loevgaard:
Fixes #97

## Problem

With autocomplete enabled, `server.search_key` is embedded in the public page source of every shop page (`AutocompleteRuntime::configuration()`), and the browser talks to Meilisearch directly. If a project misconfigures `MEILISEARCH_SEARCH_KEY` to the master key — an easy mistake, because locally both are often empty placeholders (a keyless dev instance accepts anything) — the master key is published to the world: full read/write/key-management access to the Meilisearch instance.

## Fix

Guard in `AutocompleteRuntime::configuration()`: throw when `searchKey === masterKey` and both are non-empty. Throwing at the exact point the key would be emitted **prevents the master key from ever reaching the page** — a logged warning would not stop the leak. Both-empty (keyless dev) does not trip the guard.

Also documents the hardening advice from the issue: scope the public search key to exactly the autocomplete indexes (`indexes` patterns + `actions: ["search"]`), since anything it can reach is fully queryable (prices etc. are extractable via `facetStats`/filters even when excluded from `displayedAttributes`).

## Tests

Added `AutocompleteRuntimeTest`: throws on equal keys, no throw when both empty, builds config (and does not leak the master key) when they differ. `phpunit`, `analyse`, `check-style`, `lint:container` all pass.

---

> [!NOTE]
> Final PR in the #88 → #97 stack. **Base: `issue-96-source-resolver`** (#107).

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,9 @@ setono_sylius_meilisearch:
 
 The widget talks to Meilisearch directly from the browser using the (read-only) `MEILISEARCH_SEARCH_KEY`, so make sure that key is a search-only key.
 
+> [!WARNING]
+> The search key is embedded in the public page source of every shop page and the browser queries Meilisearch directly with it. **Never** set `MEILISEARCH_SEARCH_KEY` to the master key — that would publish full read/write/key-management access to the world. The plugin guards against this: with autocomplete enabled, the container fails to compile when the search key resolves to the same non-empty value as the master key. Ideally, scope the search key to exactly the indexes used by autocomplete (Meilisearch key `indexes` patterns + `actions: ["search"]`), since anything it can reach is fully queryable — including fields hidden in the UI (prices etc. are extractable via `facetStats`/filters even when excluded from `displayedAttributes`).
+
 Because the browser connects to Meilisearch directly, it needs a URL it can actually reach. In containerized setups (Docker Swarm/Kubernetes) `MEILISEARCH_URL` is often an internal hostname like `http://meilisearch:7700` that browsers can't resolve. Configure the public, browser-facing URL separately in that case:
 
 ```yaml
--- a/src/DependencyInjection/SetonoSyliusMeilisearchExtension.php
+++ b/src/DependencyInjection/SetonoSyliusMeilisearchExtension.php
@@ -35,6 +35,7 @@ use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Exception\EnvNotFoundException;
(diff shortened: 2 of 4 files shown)


src/Resources/config/services/conditional/autocomplete.xml
```diff
@@ -16,6 +16,7 @@
             <argument type="service" id="router"/>
             <argument>%setono_sylius_meilisearch.server.public_url%</argument>
             <argument>%setono_sylius_meilisearch.server.search_key%</argument>
+            <argument>%setono_sylius_meilisearch.server.master_key%</argument>
```
loevgaard: This doesn't make sense. Why not do it in the extension before the container even compiles?

loevgaard: Good call — moved it to compile time. The check now lives in `SetonoSyliusMeilisearchExtension::registerAutocompleteConfiguration()` (only when autocomplete is enabled): it resolves `search_key` and `master_key` and throws during container compilation if they're equal and non-empty, so the misconfiguration never ships instead of failing on every page render.

The `master_key` argument on `AutocompleteRuntime` (and the runtime throw) are gone, so this PR no longer touches `autocomplete.xml` or `AutocompleteRuntime.php`. Keys that can't be resolved at compile time (injected only at runtime) skip the check rather than break the build. Force-pushed.

loevgaard: Fair point — the check is about the server keys, not autocomplete registration, so I pulled it out. It's now a standalone `assertSearchKeyIsNotMasterKey($config['server'], $config['autocomplete']['enabled'], $container)` called from `load()` right after `setServerParameters()`, and `registerAutocompleteConfiguration()` is back to its original signature (no `$serverConfig` threaded through it).

The autocomplete-enabled gate lives inside the method now, since that's the only thing that exposes the search key to the browser (and it avoids a BC break for non-autocomplete setups that happen to reuse the key). Force-pushed.

Issue #130, opened 2026-07-10
[Feature request]: Validate default_filters keys and move config assertions into Configuration


loevgaard:
## Describe the proposed solution

Two related config-surface improvements:

**1. `default_filters` keys are not validated.** The node is an untyped `scalarPrototype()` keyed by name; the recognized keys (`enabled`, `channels_aware`, `stock_available`) exist only in the extension's `array_merge` defaults. A typo like `stock_availabe: true` is silently discarded — the filter just doesn't apply, with no error. The `info()` text also only mentions the `enabled` filter.

Fix: declare the three keys as explicit `booleanNode`s (custom keys aren't supported anyway, so strictness costs nothing) and list all three in the node's `info()`.

**2. Config assertions live in the extension as bare `\RuntimeException`s** instead of `Configuration` validation, so errors don't carry config paths and surface later than they could:

- reserved index name `search`
- "when search is enabled you must configure `search.index`"
- "`search.index` must be one of the configured indexes"

Fix: move these into `validate()->ifTrue()->thenInvalid()` closures in `Configuration.php`. Also drop the contradictory `cannotBeEmpty()->defaultNull()` pair on the `indexer` node.

## Evidence

Untyped prototype:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/DependencyInjection/Configuration.php#L72-L80

Recognized keys only in the extension:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/DependencyInjection/SetonoSyliusMeilisearchExtension.php#L503-L509

Runtime exceptions that should be config validation:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/DependencyInjection/SetonoSyliusMeilisearchExtension.php#L344

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/DependencyInjection/SetonoSyliusMeilisearchExtension.php#L410-L416

Contradictory node definition:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/DependencyInjection/Configuration.php#L62-L66

## Additional context

Acceptance criteria:

[...]

PR #154, opened 2026-07-10
Let #[Facetable(sorter:)] reference a service; validate #[Sortable(direction:)]

### LINKED ISSUE #132: [Feature request]: Let #[Facetable(sorter:)] reference a service; validate #[Sortable(direction:)]
## Describe the proposed solution

Two small ergonomics fixes on the document attributes:

**1. `#[Facetable(sorter: ...)]` takes a class-string that must be instantiable with `new`** — a custom facet-value sorter cannot use dependency injection (no repositories, no config, no translator). The code itself carries a `TODO: Should be a service id`.

Fix: accept a service id, resolved through a service locator of tagged `FilterValuesSorterInterface` services (autoconfigure the interface with a tag, consistent with the plugin's other extension points). Keep plain class-string instantiation as a BC fallback when the string isn't a known service id.

**2. `#[Sortable(direction:)]` accepts any string.** `'ascending'`, `'ASC '`, or a typo are silently accepted and only misbehave later.

Fix: validate against the `ASC`/`DESC` constants in the constructor (`Assert::oneOf`), or switch to an enum.

## Evidence

The TODO and the newability assertion:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Document/Attribute/Facetable.php#L14-L23

Unvalidated direction:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Document/Attribute/Sortable.php#L12-L19

## Additional context

Acceptance criteria:

- [ ] A sorter registered as a service (with constructor dependencies) can be referenced from `#[Facetable]` by service id.
- [ ] Existing class-string usage keeps working.
- [ ] `#[Sortable(direction: 'ascending')]` throws at attribute-read time with a clear message.

loevgaard:
## What

Two ergonomics fixes on the document attributes.

**1. `#[Facetable(sorter: ...)]` can now reference a service.** Previously the sorter was a class-string instantiated with `new`, so a custom facet-value sorter could not use dependency injection. It is now resolved through a service locator of tagged `FilterValuesSorterInterface` services — the interface is autoconfigured with the `setono_sylius_meilisearch.filter_values_sorter` tag, consistent with the plugin's other extension points. Plain class-string instantiation is kept as a **BC fallback** for strings that aren't a known service id.

The shipped `SizeSorter` is now registered as a service (with its FQCN as id and the tag), so existing `#[Facetable(sorter: SizeSorter::class)]` usage keeps working — it resolves via the locator (and can now itself have dependencies).

**2. `#[Sortable(direction:)]` is validated.** The constructor now asserts the direction is one of `ASC`/`DESC` (or `null`), so `#[Sortable(direction: 'ascending')]` throws at attribute-read time with a clear message instead of silently misbehaving later.

## Testing

- `SortableTest`: allowed directions + null pass; `'ascending'` throws.
- `ChoiceFilterFormBuilderTest`: sorter resolved from the locator by id; BC class-string `new` fallback; no sorter → unsorted; unknown reference throws.
- Functional `SearchFormBuilderTest` still passes (the test app uses `sorter: SizeSorter::class`, now routed through the locator), and `lint:container` is clean.

Closes #132

---
_Stacked on #153 (#130)._

### DIFF
--- a/src/Document/Attribute/Facetable.php
+++ b/src/Document/Attribute/Facetable.php
@@ -5,21 +5,17 @@ declare(strict_types=1);
 namespace Setono\SyliusMeilisearchPlugin\Document\Attribute;
 
 use Attribute;
-use Setono\SyliusMeilisearchPlugin\Form\Builder\Sorter\FilterValuesSorterInterface;
-use Webmozart\Assert\Assert;
 
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::TARGET_METHOD)]
 final class Facetable extends Filterable
 {
     /**
-     * TODO: Should be a service id
-     *
-     * @param class-string<FilterValuesSorterInterface>|null $sorter
+     * @param string|null $sorter The facet-value sorter to apply: the service id of a service
+     *   tagged "setono_sylius_meilisearch.filter_values_sorter" (autoconfiguration adds the tag to
+     *   every Form\Builder\Sorter\FilterValuesSorterInterface implementation). The shipped
+     *   SizeSorter is registered under its FQCN, so `sorter: SizeSorter::class` works out of the box.
      */
     public function __construct(public readonly int $position = 0, public readonly ?string $sorter = null)
     {
-        if ($sorter !== null) {
-            Assert::true(is_a($sorter, FilterValuesSorterInterface::class, true));
-        }
     }
 }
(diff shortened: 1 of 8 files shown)


src/Document/Attribute/Facetable.php
```diff
@@ -5,21 +5,17 @@
...
 final class Facetable extends Filterable
 {
     /**
-     * TODO: Should be a service id
-     *
-     * @param class-string<FilterValuesSorterInterface>|null $sorter
+     * @param string|null $sorter The facet-value sorter to apply. Either the service id of a
+     *   service tagged "setono_sylius_meilisearch.filter_values_sorter" (which may have
+     *   constructor dependencies), or — for backwards compatibility — the FQCN of a
+     *   Form\Builder\Sorter\FilterValuesSorterInterface that can be instantiated with `new`.
```
loevgaard: The plugin has not been released in a stable release, so we don't want to worry about BC

loevgaard: Good call — dropped the class-string `new` fallback entirely. `#[Facetable(sorter:)]` now requires the sorter to be a registered service tagged `setono_sylius_meilisearch.filter_values_sorter` (the shipped `SizeSorter` is registered under its FQCN, so `sorter: SizeSorter::class` still works). Propagated the change through the rest of the stacked branches too.

Issue #128, opened 2026-07-10
[Feature request]: Add logging to the indexing pipeline - validation failures currently drop documents silently


loevgaard:
## Describe the proposed solution

There is no logging anywhere in the indexing pipeline. Two silent `continue`s in `DefaultIndexer::indexEntities()` — object-filter rejection and **validator rejection** — mean a document with an invalid/missing required field simply vanishes from the index with zero diagnostics. An entire catalog can fail to index and the only observable symptom is "search is empty". This is the hardest support scenario the plugin can produce, and today it's undebuggable without a step debugger.

Proposal:

- Inject a `LoggerInterface` (dedicated `setono_sylius_meilisearch` monolog channel) into `DefaultIndexer` and the message handlers.
- Log each validator rejection at `warning` with document id + violation messages; log filter rejections at `debug`.
- Track per-batch counters (mapped / filtered out / invalid / indexed) and log a summary per batch; expose the same counters so `setono:sylius-meilisearch:index` can print a summary table (see the companion IndexCommand issue).

## Evidence

The two silent drops:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Indexer/DefaultIndexer.php#L78-L84

## Additional context

Acceptance criteria:

- [ ] A document failing validation produces a `warning` log entry naming the entity id and each violation.
- [ ] A full index run logs a per-index summary (indexed vs skipped, with reasons).

PR #156, opened 2026-07-10
Remove the document from Meilisearch when an entity is deleted

### LINKED ISSUE #111: [Bug]: Deleting an entity never removes its document from Meilisearch
## Problem

Deleting an indexable entity never removes its document from Meilisearch, so deleted products stay searchable forever.

`RemoveEntity` is dispatched from the `postRemove` Doctrine event — at that point the row is already deleted. The handler then tries to **reload the entity from the database**, gets `null`, and throws `UnrecoverableMessageHandlingException`. So:

- **Async transport:** the message is marked unrecoverable and the document is never deleted — orphaned documents accumulate.
- **Sync (the default, no routing is configured):** the exception propagates out of the delete flush, so deleting an indexable entity can fail the whole request.

This is a follow-up to #68: that fix correctly captured the id in `preRemove` (via `SplObjectStorage`) so the `RemoveEntity` message carries the id, but the handler still re-`find()`s the entity, which defeats the purpose.

## Evidence

The handler reloads and throws when the entity is gone (which for a remove is the *normal* case):

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Message/Handler/AbstractEntityHandler.php#L29-L42

`RemoveEntityHandler` routes through that same `handle()`:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Message/Handler/RemoveEntityHandler.php#L13-L16

Dispatched after the row is deleted:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/EventListener/Doctrine/EntityListener.php#L45-L63

The reload is pointless anyway: the document identifier is just the stringified id, which the message already carries:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Model/IndexableAwareTrait.php#L9-L12

## Proposed fix

- Give `RemoveEntityHandler` its own implementation that does **not** reload the entity: resolve the affected indexes via `IndexRegistry::getByEntity($message->class)` and remove the document by id.
- Add a `removeByIds(array $ids): void` method to the indexer. To stay BC (there is a Roave BC-check workflow), introduce it on a new interface (e.g. `IdRemovalCapableIndexerInterface`) implemented by `DefaultIndexer`, with an `instanceof` check in the handler.
- In the new method, use the **batch** `deleteDocuments([...])` per scope instead of one `deleteDocument()` HTTP task per entity (current per-document loop: [`DefaultIndexer::removeEntities()`](https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Indexer/DefaultIndexer.php#L96-L107)).
- While in `EntityListener`: `detach()` the entity from `$removeIndexableStorage` in `postRemove` — it is currently attach-only, so the storage grows unbounded in long-running workers.

## Acceptance criteria

- [ ] Functional test: create + index a product, delete it, assert the document is gone from every scope's index.
- [ ] Deleting an indexable entity no longer throws in the default (sync) setup.
- [ ] Removal issues one `deleteDocuments` task per scope, not one task per entity.
- [ ] `EntityListener` no longer leaks `SplObjectStorage` entries after `postRemove`.

loevgaard:
## What

Deleting an indexable entity never removed its document from Meilisearch, so deleted products stayed searchable forever.

`RemoveEntity` is dispatched from `postRemove` — at that point the row is already deleted — but the handler then reloaded the entity via `find()`, got `null`, and threw `UnrecoverableMessageHandlingException`. So on async transports the message was marked unrecoverable (orphaned documents accumulate), and on the sync default the exception propagated out of the delete flush and could fail the whole request.

Changes:
- **`RemoveEntityHandler` no longer reloads the entity.** It resolves the affected indexes via `IndexRegistry::getByEntity($message->class)` and removes the document by its id — the stringified entity id the message already carries (see `IndexableAwareTrait`).
- **New `IdRemovalCapableIndexerInterface::removeByIds(array $ids)`**, implemented by `DefaultIndexer`, with an `instanceof` check in the handler so custom indexers stay BC (they're skipped with a logged warning).
- **Batch removal:** `removeByIds()` (and the refactored `removeEntities()`) issue **one `deleteDocuments` task per scope** instead of one `deleteDocument` HTTP task per entity.
- **`EntityListener::postRemove` now `detach()`es** the entity from the `SplObjectStorage`, which was attach-only and grew unbounded in long-running workers.

## Testing

- `RemoveEntityHandlerTest`: removes by id **without reloading** the entity (so it no longer throws in the sync setup); skips indexers that can't remove by id.
- `DefaultIndexerTest::it_removes_documents_with_one_batch_delete_task_per_scope`: a single `deleteDocuments([...])` per scope.
- `EntityListenerTest`: after `preRemove` + `postRemove`, a `RemoveEntity` is dispatched and the `SplObjectStorage` retains no entries.

Closes #111

---
_Stacked on #155 (#128)._


src/Indexer/IdRemovalCapableIndexerInterface.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Setono\SyliusMeilisearchPlugin\Indexer;
+
+/**
+ * Indexers implementing this interface can remove documents by their identifier alone, without
+ * having to load the underlying entity. This is what makes deletion work: when an entity is
+ * removed, the entity is already gone from the database, so the only thing left to go on is the
+ * document identifier (the stringified entity id).
+ */
+interface IdRemovalCapableIndexerInterface
```
loevgaard: We don't care about BC, so shouldn't this just be default functionality instead of an interface?

loevgaard: Agreed — removed the `IdRemovalCapableIndexerInterface` entirely and put `removeByIds()` straight on `IndexerInterface`. `RemoveEntityHandler` now just calls `$index->indexer()->removeByIds([...])` with no `instanceof` guard. Propagated through the stacked branches.

Issue #127, opened 2026-07-10
[Feature request]: Only run per-facet sub-queries for facets with active selections


loevgaard:
## Describe the proposed solution

Every search currently runs **one extra Meilisearch sub-query per facetable attribute** (each re-running the whole search with all filters except its own, `limit 1`) plus the main query, inside a single `multiSearch`. That's correct disjunctive faceting, but the cost scales linearly with the number of facets and is paid on every page load, keystroke-search, filter change, and pagination click.

The optimization: a dedicated sub-query is only *needed* for facets the user has **actively filtered on** (because only there must the facet's own filter be excluded to show the full distribution). All other facets can read their counts from the main query's `facetDistribution` for free.

Typical result: N+1 queries → 1 query with no filters active, 2–3 with one or two facets selected.

## Evidence

One sub-query per facet, unconditionally:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Meilisearch/Query/MultiSearchBuilder.php#L62-L81

## Additional context

Acceptance criteria:

- [ ] With no filters selected, exactly one search query is issued and all facet counts are populated.
- [ ] With a filter on facet X, facet X's distribution still shows the counts it would have *without* its own filter (disjunctive behaviour preserved — cover with a functional test).

Issue #72, opened 2025-03-27
[Feature request]: Manage situations where the index is not yet created


loevgaard:
### Describe the proposed solution

It gives an exception when the index is not yet created and you try to search. Instead we could handle this by catching the exception and checking if it was thrown because the index doesn't exist and then render a page where it explains that to the user

PR #162, opened 2026-07-10
Do not register the search routes when search is disabled

### LINKED ISSUE #129: [Feature request]: Do not register the search controller/route when search is disabled
## Describe the proposed solution

The search controller and its service wiring are registered even when `setono_sylius_meilisearch.search.enabled` is `false` — only a runtime parameter guards the behaviour. This is an acknowledged todo in the extension. When search is disabled, the `SearchAction` service (and anything only it needs) shouldn't exist, and the shop route shouldn't be reachable.

## Evidence

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/DependencyInjection/SetonoSyliusMeilisearchExtension.php#L394-L395

```php
/**
 * todo the search controller should only be available when search is enabled
 */
```

## Describe alternatives you've considered

Keeping the runtime guard but returning a 404 from the action — works, but leaves dead services in the container and the route in the router.

## Additional context

Related: #72 tracks the adjacent first-run problem (search enabled but the index not created yet → raw Meilisearch `ApiException`/500). Worth solving together: when search **is** enabled but the index is missing, catch the `ApiException` and either render the no-results state with a logged error (prod) or throw a descriptive exception telling the developer to run `setono:sylius-meilisearch:index` (dev).

Acceptance criteria:

- [ ] With `search.enabled: false`, the search route is not registered and `SearchAction` is not in the container.
- [ ] (If bundled with #72) with search enabled and no index created, dev mode shows an actionable message naming the missing uid and the index command.

loevgaard:
## What

The search controller and its wiring already only exist when `search.enabled` (they live in `services/conditional/search.xml`, loaded only when enabled — the extension test already asserts `SearchAction` is absent when disabled). But the shop search **routes** were always registered, pointing at non-existent controllers when search was off — the acknowledged todo in the extension.

The shop routes are now loaded through a custom `SearchRouteLoader` (tagged `routing.loader`) that returns them only when search is enabled. The route YAMLs import the shop routes via this loader (`type: setono_sylius_meilisearch_search`) instead of importing `shop.yaml` directly. The admin synonym routes are unaffected.

## Testing

- `SearchRouteLoaderTest`: registers no routes when disabled; `supports()` only matches its own type.
- With search enabled (the test app), `debug:router` shows the three shop routes and the functional `SearchSortingTest`/`SearchPaginationTest` (which hit the search route over HTTP) still pass.

> The related first-run problem (#72 — search enabled but index not yet created) is left for that issue.

Closes #129

---
_Stacked on #161 (#127)._


src/Routing/SearchRouteLoader.php
```diff
@@ -0,0 +1,39 @@
...
+        parent::__construct($env);
+    }
+
+    public function load(mixed $resource, ?string $type = null): RouteCollection
+    {
+        $routes = new RouteCollection();
+
+        if ($this->enabled) {
+            /** @var RouteCollection $imported */
+            $imported = $this->import('@SetonoSyliusMeilisearchPlugin/Resources/config/routes/shop.yaml');
```
loevgaard: This is wrong. What if I accidentally add routes to shop.yaml that isn't related to search?

loevgaard: Good catch. Renamed the gated file `routes/shop.yaml` → `routes/search.yaml` so it clearly holds only the search routes; the loader now imports `search.yaml`, and the `routes.yaml`/`routes_no_locale.yaml` import keys + comments were updated to say "search" too. Now a generic shop route can never be swept up by the `search.enabled` toggle. Public route names are unchanged (`setono_sylius_meilisearch_shop_search` etc.), and `debug:router` confirms they still register when enabled.

Issue #134, opened 2026-07-10
[Feature request]: i18n gaps on the search page - facet-label fallback, hardcoded strings, pluralization


loevgaard:
## Describe the proposed solution

Several shopper-visible strings on the search page don't localize:

1. **Facet headings fall back to the raw translation key.** Labels are built as `setono_sylius_meilisearch.form.search.facet.<name>`, but only `color`, `on_sale`, `price`, `size`, `taxons` ship. Any other facet — e.g. `brand`, which the plugin's own e2e suite uses — renders the literal key string as its heading. Fix: when the translator catalogue doesn't define the key, fall back to a humanized facet name (`ucfirst(str_replace('_', ' ', $name))`); keep the key lookup so projects can override.
2. **Hardcoded English:** "No results found" / "Try to search for something else" (`_no_results.html.twig`) and the `Min` / `Max` range labels (`RangeType`). Move to translation keys.
3. **No pluralization:** `{{ searchResult.totalHits }} results` renders "1 results" (the e2e even asserts it). Use an ICU-pluralized key. The `on_sale` label's `(%count%)` suffix has the same issue.

All new keys should be translated into the 11 existing locales (keys sorted alphabetically, per repo convention), and the `"1 results"` e2e assertion updated.

## Evidence

Label key construction (same pattern in all three builders):

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Form/Builder/ChoiceFilterFormBuilder.php#L27

Shipped facet keys:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Resources/translations/messages.en.yaml#L4-L9

Hardcoded strings:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Resources/views/search/_no_results.html.twig#L1-L6

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Resources/views/search/_hits_count.html.twig#L1-L2

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Form/Type/RangeType.php#L16-L24

## Additional context

Acceptance criteria:

[...]

PR #167, opened 2026-07-10
Add a submit button so the search form works without JavaScript

### LINKED ISSUE #122: [Bug]: The search results form has no submit button - filters, sorting and pagination do not work without JavaScript
## Problem

The search results form (facets + sorting + pagination) renders `form_start` … `form_end` with **no submit button anywhere**. All submission is done by the JavaScript auto-submit in `search.js`. With JavaScript disabled, blocked, or failed to load, a shopper can run the initial query from the URL but cannot apply a filter, change sorting, or paginate — the form is dead.

This contradicts `search.js`'s own header comment ("search keeps working even without JavaScript"), which is only true for the plain query submit, and it's a hard blocker for no-JS clients and crawlers following the filter UI.

## Evidence

The full form — no submit control:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Resources/views/search/_results.html.twig#L1-L23

The progressive-enhancement claim:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Resources/public/js/search.js#L1-L11

JS is the only thing submitting the form:

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Resources/public/js/search.js#L196-L221

## Proposed fix

Add a real submit button to `_results.html.twig` (e.g. "Apply filters"), visible by default and hidden by `search.js` on init (not via `<noscript>` alone — that doesn't cover failed/blocked JS). The GET form then works everywhere; JS keeps the enhanced auto-submit behaviour.

## Acceptance criteria

- [ ] With JS disabled (Playwright `javaScriptEnabled: false`), checking a facet and pressing the button filters the results; sort and pagination work the same way.
- [ ] With JS enabled, the button is not visible and behaviour is unchanged.

loevgaard:
## What

The search results form (facets + sorting + pagination) rendered `form_start … form_end` with **no submit button anywhere** — all submission was done by the JavaScript auto-submit in `search.js`. With JavaScript disabled, blocked, or failed to load, a shopper could run the initial query from the URL but couldn't apply a filter, change sorting, or paginate.

Add a real **"Apply filters"** submit button to `_results.html.twig`, visible by default (so the plain GET form works everywhere) and hidden by `search.js` on init (so the enhanced auto-submit behaviour is unchanged when JS is available). Hiding it in JS — rather than via `<noscript>` — also covers failed/blocked JS.

The `apply_filters` key is translated into all 12 locales.

## Testing

- e2e `filters by a brand facet`: with JS enabled the fallback button is hidden and checking a facet auto-submits.
- e2e `without javascript` (`javaScriptEnabled: false`): the button is visible; checking a facet and pressing it filters the results.
- The functional search-page tests still pass (template renders with the button + translation key).

Closes #122

---
_Stacked on #166 (#134)._

### DIFF
--- a/src/Resources/public/js/search.js
+++ b/src/Resources/public/js/search.js
@@ -193,6 +193,15 @@ class SearchManager {
 
         this.#form = form;
 
+        // JavaScript is available, so changing a filter/sort/page auto-submits. Hide the
+        // fallback submit button, which only exists so the form works without JavaScript.
+        // Set display directly as well: the button carries Semantic UI's `.ui.button`
+        // classes, whose `display` rule outweighs the `[hidden]` attribute on its own.
+        this.#form.querySelectorAll('.ssm-apply-filters').forEach((el) => {
+            el.hidden = true;
+            el.style.display = 'none';
+        });
+
         this.#form.addEventListener('input', (event) => {
             const field = event.target;
 
--- a/src/Resources/translations/messages.da.yaml
+++ b/src/Resources/translations/messages.da.yaml
@@ -1,6 +1,7 @@
 setono_sylius_meilisearch:
     form:
         search:
+            apply_filters: 'Anvend filtre'
             facet:
                 color: Color
                 on_sale: On sale (%count%)
--- a/src/Resources/translations/messages.de.yaml
+++ b/src/Resources/translations/messages.de.yaml
@@ -1,6 +1,7 @@
 setono_sylius_meilisearch:
     form:
         search:
+            apply_filters: 'Filter anwenden'
             facet:
                 color: Color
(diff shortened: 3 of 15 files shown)


loevgaard:
Closing this for now — the fallback submit button flashes visible for a moment before JS hides it, which is bad UX. Keeping the underlying issue #122 open to revisit with a better approach (e.g. hiding it via CSS by default and only revealing it when JS is absent, so there's no flash). The stacked PRs above have been rebased to drop this commit.

Issue #140, opened 2026-07-10
[Docs]: Document the operational model - async indexing, settings sync, production checklist


loevgaard:
## Describe the proposed solution

The README covers installation and attribute usage but says nothing about how the plugin behaves operationally — the words "messenger", "settings", "downtime" and "deploy" don't appear at all. Three sections are missing:

**1. Asynchronous indexing (strongly recommended).** Out of the box every entity save indexes synchronously inside the request (see the companion resilience issue). Document the exact `framework.messenger.routing` snippet for routing the plugin's commands (they live on the dedicated `setono_sylius_meilisearch.command_bus`, which makes the routing non-obvious), plus `messenger:consume` supervision and failure-transport guidance.

**2. How settings sync works.** Meilisearch index settings (`filterableAttributes` from `#[Facetable]`, `sortableAttributes` from `#[Sortable]`, …) are only pushed by the full `setono:sylius-meilisearch:index` command — incremental Doctrine-event indexing never updates settings. A developer who adds `#[Facetable]` to a document property and waits for auto-indexing gets a silently non-functional facet. One prominent callout ("after changing document attributes, run the index command") saves hours of confusion.

**3. Production checklist.** Run the index command on deploy; nightly reindex cron example (one line, with `--wait` note); worker supervision; which env vars are required vs optional (`MEILISEARCH_URL`/`MASTER_KEY` required; `SEARCH_KEY` needed for search/autocomplete; `PREFIX`/`PUBLIC_URL` optional — `MEILISEARCH_PUBLIC_URL` is currently missing from the install env block entirely and only appears in the autocomplete section).

Also: link the test application's config as the worked reference example — it's the best existing documentation of a multi-index setup and is currently only discoverable by reading the test suite.

## Evidence

- Settings only updated in the full-index handler: https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Message/Handler/IndexHandler.php#L36-L52
- Reference config never linked from docs: https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/tests/Application/config/packages/setono_sylius_meilisearch.yaml

## Additional context

Acceptance criteria:

- [ ] README gains the three sections above.
- [ ] Every documented config key/command is verified against the current source.

Issue #131, opened 2026-07-10
[Feature request]: Improve IndexCommand output - resolved uids, result summary, honest --delete warning


loevgaard:
## Describe the proposed solution

`setono:sylius-meilisearch:index` prints only `Index "%s" submitted for indexing` and returns. Two problems:

- In the default (synchronous) setup, "submitted" is misleading: the command appears to hang while it is actually doing all the indexing inline, with no progress signal for large catalogs. With async transports it returns instantly with no way to observe the result besides `--wait`.
- There is no summary: how many entities were indexed, skipped by filters, or dropped by validation.

Proposal:

- Print the resolved scopes/uids per index up front (`Indexing products → products__fashion_web__en_US (…)`).
- With the pipeline counters from the logging issue, print a per-index summary table (indexed / filtered / invalid) at the end when running synchronously, and after `--wait` when waiting.
- Reword the `--delete` warning to state the consequence explicitly ("the live index is deleted first — search returns no results until reindexing completes"), until a swap-based rebuild makes the flag obsolete.

## Evidence

https://github.com/Setono/sylius-meilisearch-plugin/blob/14bef062bd4ed690139c3748179c159e854d0a7b/src/Command/IndexCommand.php#L82-L91

## Additional context

Acceptance criteria:

- [ ] Command output names each resolved index uid.
- [ ] A summary with counts is printed (sync run or after `--wait`).
- [ ] The `--delete` warning explains the downtime.

PR #186, opened 2026-08-20
Remove the persistent document metadata cache

### LINKED ISSUE #185: [Bug]: Cached document metadata desyncs index settings from indexed documents - new #[Searchable] fields silently never become searchable
### Version

dev-master (bc05a89), Symfony 6.4, default plugin config (`metadata.cache` defaults to `!kernel.debug`)

### What happens

`DefaultIndexer` normalizes documents with the Symfony serializer, so indexed documents always reflect the **live document class**. `IndexHandler` pushes index settings (`searchableAttributes`, `filterableAttributes`, …) built from `MetadataFactoryInterface`, which in non-debug kernels is the `CachedMetadataFactory` — a cache with **no invalidation tied to the document class definition**.

After editing a document class (e.g. adding a `#[Searchable]` property), a reindex in any environment where the metadata cache is warm therefore produces a **silent desync**:

- the documents in Meilisearch contain the new field ✅
- the `settingsUpdate` task still pushes the old attribute lists ❌

so the new field never becomes searchable/filterable/facetable. No error, no log line, the index command reports success. The only way to notice is to inspect the index settings via the Meilisearch API.

### Reproduction

1. Run `setono:sylius-meilisearch:index` once with `APP_DEBUG=0` (metadata gets cached).
2. Add `#[Searchable] public ?string $code = null;` to your document + a data mapper that fills it.
3. Run `setono:sylius-meilisearch:index` again with `APP_DEBUG=0` (same `var/cache`, no `cache:clear` in between).
4. `GET /indexes/<uid>/settings` → `searchableAttributes` unchanged, while `GET /indexes/<uid>/documents` shows the new field on every document. Searching the new field returns nothing.

### What makes it extra confusing

The pool (`setono_sylius_meilisearch.cache.metadata`, `cache.system` adapter) is namespaced by the compiled container class, which includes the debug flag. So `cache:pool:clear setono_sylius_meilisearch.cache.metadata` run in a debug-mode process **does not clear what a non-debug indexer reads**. A very common workflow — developer runs `cache:clear`/pool clears in debug mode, but runs indexing with `APP_DEBUG=0` to keep the dev profiler from exhausting memory on a full catalog — stays stale indefinitely. This cost us several hours while shipping a "make product numbers searchable" fix: documents kept gaining the field while search kept ignoring it.

### Expected

Settings pushed by the index run should match the document class the very same run just serialized — or at minimum the staleness should be visible.

### Suggested fixes (any of, roughly in order of preference)

1. **Fingerprint-based invalidation**: include a hash of the document class definition in the cache item (e.g. `filemtime`/`ReflectionClass::getFileName()` over the class hierarchy, or a hash of the reflected attribute config) and rebuild on mismatch. Keeps the cache, removes the trap.
2. **Make staleness visible**: have the index run log/print the attribute lists it pushes per index — dovetails with #131 (IndexCommand output improvements). A stale list next to the uid would have made this a 2-minute diagnosis.
3. **Document the pool**: name the pool, when it is populated, and that deployments/long-lived processes must clear it after document changes — dovetails with #140 (operational model docs).

Happy to PR whichever direction you prefer.

loevgaard:
Fixes #185

## The problem

The metadata cache pool had no invalidation tied to the document class definition. After editing a document class (e.g. adding a `#[Searchable]` property), a reindex with a warm cache produced a silent desync: documents in Meilisearch contained the new field, but the `settingsUpdate` task still pushed the old attribute lists, so the field never became searchable/filterable/facetable. Clearing the pool from a debug-mode process didn't help either, since the pool namespace differs between debug and non-debug containers.

## Why removal instead of fixing the invalidation

An earlier revision of this PR added fingerprint-based invalidation (file mtimes over the class hierarchy), but measuring showed the cache isn't worth its risk. On the real `Product` document:

| Operation | Cost |
|---|---|
| Full metadata rebuild (reflection + attribute instantiation) | ~9 µs |
| Pool hit, lower bound (`unserialize` alone, excluding PSR-6 overhead) | ~3 µs |

The pool saves at most ~6 µs **once per request per document class**, and only on requests that touch metadata at all (search/taxon pages) - noise next to the Meilisearch HTTP round trip. Meanwhile mtime-based invalidation has its own corner cases (reproducible-build images that clamp mtimes, `opcache.validate_timestamps=0` with in-place deploys). Deleting the cache makes the #185 bug class impossible instead of mitigated.

## What changed

- `CachedMetadataFactory` and its test are deleted, along with the `metadata.cache` config node, the `setono_sylius_meilisearch.cache` parameter, the prepended `setono_sylius_meilisearch.cache.metadata` pool, and the `kernel.debug` plumbing in `Configuration`. (No release has ever been tagged, so removing the config node outright is safe; anyone who set `metadata.cache` explicitly gets a loud "Unrecognized option".)
- The in-memory memoization moves into `MetadataFactory` itself. This is what the hot path actually needs: data mappers call `getMetadataFor()` twice per entity during indexing, and indexing runs in long-lived processes (messenger workers, the CLI command), so the ~9 µs build happens once per process. In-process memoization is inherently consistent - a process cannot see two versions of a class - so settings always derive from the exact class the same run serialized, which is the "Expected" line from #185 by construction.
- `MetadataCreated` is now dispatched once per document class per process (previously once per pool lifetime when the cache was warm, or on every call when it wasn't).

## Tests

`MetadataFactoryTest` gains `it_memoizes_metadata` (same instance returned for repeated calls with the class name and with an instance, event dispatched once). `composer fix-style`, `composer analyse`, `composer phpunit`, `vendor/bin/rector process --dry-run`, and `lint:container` all pass.

The other suggestions from the issue (index-run settings output, operational docs) remain for #131 / #140.


src/DependencyInjection/SetonoSyliusMeilisearchExtension.php
```diff
@@ -109,19 +99,12 @@ public function load(array $configs, ContainerBuilder $container): void
 
     public function getConfiguration(array $config, ContainerBuilder $container): Configuration
     {
-        return new Configuration((bool) $container->getParameter('kernel.debug'));
+        return new Configuration();
     }
```
loevgaard: This can be removed totally now, right?

loevgaard: Yes — removed in 2da8584. The only wrinkle was that the base `Extension::getConfiguration()` returns `?ConfigurationInterface`, which PHPStan flags when passed to `processConfiguration()`, so `load()` now does `processConfiguration(new Configuration(), $configs)` directly (and the `@psalm-suppress PossiblyNullArgument` became obsolete too). `config:dump-reference` still resolves `Configuration` via the base class's naming convention.

PR #190, opened 2026-08-20
Let store admins configure which product attributes and options are indexed

### LINKED ISSUE #187: [Feature]: Let store admins choose which product attributes and options are indexed (searchable/filterable/facetable)
## Motivation

Product attributes and options are merchant-defined data, but *which* of them end up in the Meilisearch index is developer-defined code: document classes hardcode the codes in `#[MapProductAttribute([...])]` / `#[MapProductOption([...])]` and the roles via `#[Searchable]` / `#[Facetable]` / `#[Filterable]` / `#[Sortable]`. When a merchant adds a new attribute (say "Material"), making it facetable requires a code change and a deploy — the exact workflow that produced #185. Admin-configurable indexing ("check off attributes in the backend") closes that gap and matches what merchants expect from comparable integrations (e.g. Algolia's Magento facet configuration).

## Sketch

- A new resource, e.g. `IndexableAttribute` (attribute or option code, target index, flags: searchable, filterable, facetable with position + optional sorter, sortable), managed via a Sylius admin grid — same pattern as the existing `Synonym` resource.
- A `MetadataCreated` listener merges the persisted configuration into the `Metadata` built from the document class. The event already exists precisely as this kind of extension point, and everything downstream — `DefaultSettingsProvider`/`IndexHandler` on the settings side, `SearchFormBuilder` and the facet machinery on the search side — reads `Metadata`, so dynamic attributes flow through without touching those layers.
- Documents need a dynamic value bag for admin-selected attributes; `AttributesDataMapper` / `DocumentAttributesValuesSetter` already resolve localized attribute values, but today only into properties declared via PHP attributes.
- Facet type should derive from the Sylius `AttributeType` (select/text → string list, integer/percent → number, checkbox → bool) instead of reflected property types.
- Saving the configuration must orchestrate a settings update **and** a reindex together (dispatch the existing messenger command, or surface a "reindex required" notice in the admin). Documents and settings changing in lockstep is the lesson of #185.

## Metadata caching implications (from the #186 discussion)

- With a DB-driven part, building metadata is no longer ~9µs of pure reflection (the measurement that justified removing the cache in #186) but reflection + a config query. Still small, but no longer trivially negligible.
- The in-process memoization in `MetadataFactory` gets a new failure mode: a long-lived messenger worker would keep indexing with stale admin config until restart. `MetadataFactory` should implement `ResetInterface` (clearing the memoized metadata; messenger resets services between messages), so each message sees fresh config at a cost of one rebuild per message.
- If a shared cache is wanted for web requests, it should cache the *config lookup*, keyed by a version/`updatedAt` of the configuration and invalidated on save. Runtime config changes flow through the application, so invalidation is deterministic — unlike the file-edit invalidation problem that killed the old reflection cache (#185). Probably unnecessary in a first iteration: the uncached cost is one small indexed query per request, and only on pages that touch metadata.

loevgaard:
Fixes #187

## What this does

Adds two admin screens (**Meilisearch → Indexed attributes** and **Meilisearch → Indexed options**, mirroring the Catalog section's attribute/option split) where the shop admin picks product attributes and product options, the indexes they apply to, and their roles — **searchable**, **filterable**, **facetable** — without code changes or deploys. Saving automatically updates the affected indexes' settings and reindexes their documents, so the two can never desync (the #185 lesson).

## How it works

- **Index-aware metadata** — `MetadataFactoryInterface::getMetadataFor()` now takes an optional `Config\Index`; memoization is keyed per class+index and `MetadataCreated` carries the index. This is what makes per-index configuration possible (e.g. attributes in the real products index but not in a products-autocomplete index). `MetadataFactory` implements `ResetInterface` with an explicit `kernel.reset` tag, so long-running messenger workers see configuration changes between messages.
- **Two resources, one shared base** — `IndexableAttribute` and `IndexableOption` are separate resources (own tables, repositories, grids, CRUD routes, forms, menu entries) sharing the abstract `IndexableSubject` base for the common state (code, indexes, roles, facet position). Attributes and options are distinct concepts in Sylius, and the split also keeps every form a plain mapped `code` choice field — no type discriminator, no encoded subject values. Validation: `UniqueEntity` on the code, at least one role, at least one index.
- **DB-driven metadata** — a `MetadataCreated` subscriber merges enabled rows from both repositories into the metadata as flat `attr_<code>`/`opt_<code>` fields (form-safe, filter-expression-safe, collision-proof against declared document fields). It enforces the facetable ⊆ filterable invariant (`MultiSearchBuilder` requests every facetable name on every search; a name missing from Meilisearch's `filterableAttributes` fails the request) and coerces storage types: select/text → `array`, checkbox → `bool`, integer/float → `float` (there is no `int` filter builder), date/datetime → searchable/filterable only, never facets. Deleted attributes, unsafe codes, and name collisions degrade to logged skips.
- **Indexing** — `DynamicFieldsDataMapper` reuses the existing `AttributesDataMapperValuesProvider`/`OptionsDataMapperValuesProvider` (localization comes free) to fill a new `$dynamicFields` bag on `Document`; `DocumentNormalizer` flattens the bag to top-level fields at serialization time, with declared fields always winning.
- **Facet labels** — `FacetLabelTrait` is replaced by an injected `FacetLabelResolver`: translation catalogue key first (so projects can override), then the Sylius attribute's/option's translated name for the current locale, then the humanized fallback.
- **Eligible indexes only** — the admin can only target indexes that index products or product variants and that have not opted out via the per-index `dynamic_fields: false` configuration flag (e.g. to keep an autocomplete index lean). Ineligible indexes are hidden from the index choices, and the metadata subscriber ignores stale rows referencing them. Variant sources are fully supported: a variant document gets its parent product's attributes and its own option values.
- **Save → sync** — a Doctrine listener watches both resources, collects affected index names (including indexes *removed* from a row, via the `preUpdate` changeset), resets the metadata memo, and dispatches one `Index` command per index on kernel response — mirroring `SynonymListener`. Between the save and Meilisearch processing the settings task, `SearchEngine` catches `invalid_search_facets` and retries once without facets, so the shop degrades to a facet-less page instead of erroring; this also protects code-level `#[Facetable]` additions deployed before a reindex.

## Verification

- ECS, PHPStan (level max), Rector, `composer validate`/`normalize`, `lint:container`, `doctrine:schema:validate` all pass.
- 144 unit tests (13 new/updated covering the subscriber, mapper, normalizer, listener, label resolver, per-index memoization).
- 20 functional tests pass, run both without and with dynamic facets live — including repository index-prefix safety and an end-to-end metadata test persisting rows of both resources against the shipped fixture attributes.
- Manual end-to-end smoke against a real Meilisearch: rows for the `t_shirt_material` attribute and `t_shirt_size` option produced correct settings, and searching **"viscose"** returned exactly the 2 products whose only match is the dynamic `attr_t_shirt_material` field — proving the searchable role — alongside correct facet distributions (`"100% viscose": 2`, sizes S–XXL with counts).

## Known limitations (documented in the README)

No sortable role (v1 scope); a row's roles apply to every index it targets; numeric-valued select facets are skipped by the choice widget; date/datetime attributes cannot be facets; CLI/fixture edits don't auto-reindex; uniqueness of the code is validator-enforced only (mapped superclasses cannot declare unique constraints — the README suggests adding DB indexes in the project migration).

One deviation from the reviewed plan (besides the attribute/option split, discussed in the conversation): `DynamicFieldsDataMapper` depends on `DataMapperValuesProviderInterface` instead of the two final provider classes (final classes cannot be prophesized, and the interface is the cleaner contract; the service wiring still injects the two concrete providers).


src/Document/Document.php
```diff
@@ -30,6 +30,15 @@ abstract class Document
...
     public ?string $entityClass = null;
 
+    /**
+     * Values of fields that are added at runtime (e.g. admin-configured product attributes/options),
+     * indexed by the field name. The bag is flattened to top level fields when the document is
+     * normalized, so every key ends up as its own field in Meilisearch
+     *
+     * @var array<string, bool|float|string|list<string>>
+     */
+    public array $attributes = [];
```
loevgaard: Should this be named something that encompasses both options and attributes?

loevgaard: Agreed — renamed to `$dynamicFields` in b474616. It matches the existing naming family (`DynamicField`, `Metadata::$dynamicFields`, `DynamicFieldsDataMapper`), and since `DocumentNormalizer` flattens the bag away before the document reaches Meilisearch, the key is purely internal — no indexed data changes shape.

